PR #10, opened 2020-09-22
Separate build related properties from other properties

talarian1:
Because build properties (i.e Build name, number and VCS url) shouldn't be set with
multiple values it was decided to pass them separately.
Fixes https://github.com/jfrog/jfrog-cli/issues/786

### DIFF
--- a/artifactory/commands/generic/upload.go
+++ b/artifactory/commands/generic/upload.go
@@ -99,7 +99,7 @@ func (uc *UploadCommand) upload() error {
 		if err := utils.SaveBuildGeneralDetails(uc.buildConfiguration.BuildName, uc.buildConfiguration.BuildNumber); err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


artifactory/commands/generic/upload.go
```diff
@@ -178,30 +179,23 @@ func getMinChecksumDeploySize() (int64, error) {
 	return minSize * 1000, nil
 }
 
-func addBuildProps(props *string, buildName, buildNumber string) error {
-	if buildName == "" || buildNumber == "" {
-		return nil
-	}
+func addBuildProps(buildName, buildNumber string) (props string, err error) {
```
eyalbe4, line 182: The name of this function is no longer accurate, because the function no longer "adds" build props.
Actually, thus function has become completely redundant - you can just remove it, and call the API it includes directly.

eyalbe4 (changes requested):

PR #20, opened 2020-10-13
Add refreshable tokens exclusion for default server

RobiNino:
Refreshable tokens are disabled when working with external tools (build tools, curl, etc) or when sending requests not via ArtifactoryHttpClient.
This PR adds a fix to disable refreshable tokens both when using and the default server, and when a specified one.


eyalbe4 (approved):
For future traceability, let's add to this PR's description the actual issue it solves.

PR #51, opened 2020-12-14
Support 'include-deps' and 'exclude-artifacts' flags for search-based commands

barbelity:
-----


eyalbe4 (changes requested):
```exclude-artifacts``` should be changed to ```deps-only```. Let's change all variables, functions and messages accordingly.

barbelity:
We discussed about changing the 'exclude-artifacts' flag to 'deps-only', and decided to leave it as 'exclude-artifacts'.
Therefore, I am marking as 'resolved' all the comments requesting this change.

PR #62, opened 2021-01-27
Allow running pip install without setup.py

yahavi:
-----

Fixes https://github.com/jfrog/jfrog-cli/issues/948

This PR resolves 2 issues:
1. When setup.py does not exist in CWD, running `jfrog rt pipi <package-name> --build-name=<buildName> --build-number=<buildNumber>` will fail with "Could not find setup.py file in current directory". Only when the `-r` flag provided, the execution didn't fail. The preferred behavior should be: If setup.py does not exist, always use the build name as the module name.
2. When setup.py does exist in CWD, running `jfrog rt pipi <package-name> --build-name=<buildName> --build-number=<buildNumber> -r requirements.txt` will set the module name as the build-name, although it has a setup.py file.

### DIFF
--- a/artifactory/commands/pip/install.go
+++ b/artifactory/commands/pip/install.go
@@ -156,25 +156,16 @@ func (pic *PipInstallCommand) prepare() (pythonExecutablePath string, err error)
 }
 
 func getPackageName(pythonExecutablePath string, pipArgs []string) (string, error) {
-	// Check if using requirements file.
-	isRequirementsFileUsed, err := isCommandUsesRequirementsFile(pipArgs)
-	if err != nil {
-		return "", err
-	}
-	if isRequirementsFileUsed {
-		return "", nil
-	}
-
 	// Build uses setup.py file.
 	// Setup.py should be in current dir.
-	filePath, err := getSetuppyFilePath()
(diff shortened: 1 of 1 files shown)


artifactory/commands/pip/install.go
```diff
@@ -156,25 +156,16 @@ func (pic *PipInstallCommand) prepare() (pythonExecutablePath string, err error)
...
 	if err != nil {
 		return "", err
 	}
 
 	if filePath == "" {
-		// Couldn't resolve requirements file or setup.py.
-		return "", errorutils.CheckError(errors.New("Could not find installation file for pip command, the command must include '--requirement' or be executed from within the directory containing the 'setup.py' file."))
+		// setup.py does not exist in directory.
+		return "", nil
 	}
```
eyalbe4, line 169: We can reduce the following code:
```
	if err != nil {
		return "", err
	}

	if filePath == "" {
		// setup.py does not exist in directory.
		return "", nil
	}
```
into:
```
         // Error was returned or setup.py does not exist in directory.
	if err != nil || filePath == "" {
		return filePath, err
	}
```

eyalbe4 (changes requested):

PR #103, opened 2021-04-01
Fix buildinfo agent

asafgabai:
-----


utils/coreutils/utils.go
```diff
@@ -22,13 +22,13 @@ type OnError string
...
-var cliUserAgent string
+// User agent
+var cliUserAgentName string
+var cliUserAgentVersion string
 
-var version string
-
-// Name of the agent, i.e 'jfrog-cli-go'.
-var clientAgent string
+// Client agent
```
eyalbe4, line 29: It's not really clear to the random reader what  "Client agent" means. What are the values that it holds and what it is used for.
Let's elaborate more on the above "User agent" values as well.

eyalbe4 (changes requested):

PR #102, opened 2021-04-01
Download build-info extractors from ojo

yahavi:
- [x] All [tests](https://github.com/jfrog/jfrog-cli-core#tests) passed. If this feature is not already covered by the tests, I added new tests.
- [x] This pull request is on the dev branch.
- [x] I used gofmt for formatting the code before submitting the pull request.
-----

Following Bintray sunset, we should download Maven and Gradle build-info extractors JARs from `https://oss.jfrog.org/artifactory` instead of from JCenter.

This PR introduce the following changes:
1. Introduce `JFROG_CLI_EXTRACTORS_REMOTE`. This env var in form of ServerID/RemoteRepo.
2. Deprecate `JFROG_CLI_JCENTER_REMOTE_SERVER` and suggest to use `JFROG_CLI_EXTRACTORS_REMOTE` instead.
3.  * If `JFROG_CLI_JCENTER_REMOTE_SERVER` and `JFROG_CLI_EXTRACTORS_REMOTE` are not set, download JARs from  https://oss.jfrog.org/artifactory.
     * If `JFROG_CLI_JCENTER_REMOTE_SERVER` is set - use it and `JFROG_CLI_JCENTER_REMOTE_REPO` as before and print a warning message. Following this error message, it is the user's responsibility to change the remote repository URL from https://jcenter.bintray.com to https://oss.jfrog.org.
    * If `JFROG_CLI_EXTRACTORS_REMOTE` is set, download the JARs from the configured server ID and repository.
4. Update Maven and Gradle extractors version to latest.

PR #105, opened 2021-04-02
Use different extractors version when using JCenter

yahavi:
-----
Thie PR is a follow-up to #102.
Following JCenter sunset, we stopped publishing the build-info extractors. Therefore, the version in JCenter is not the latest.

If `JFROG_CLI_JCENTER_REMOTE_SERVER` deprecated env var is set, use Maven extractor 2.23.0 instead of 2.25.0 and Gradle extractor 4.21.0 instead of 4.23.0.

### DIFF
--- a/artifactory/commands/mvn/mvn.go
+++ b/artifactory/commands/mvn/mvn.go
@@ -21,6 +21,7 @@ import (
 )
 
 const mavenExtractorDependencyVersion = "2.25.0"
+const mavenExtractorDependencyJCenterVersion = "2.23.0"
 const classworldsConfFileName = "classworlds.conf"
 const MavenHome = "M2_HOME"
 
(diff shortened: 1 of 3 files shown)


artifactory/commands/mvn/mvn.go
```diff
@@ -21,6 +21,7 @@ import (
 )
 
 const mavenExtractorDependencyVersion = "2.25.0"
+const mavenExtractorDependencyJCenterVersion = "2.23.0"
```
eyalbe4, line 24: Let's a descreiption for this constant. It is best to include the word "deprecated" to the description, so that we remember to remove it in jfrog-cli 2.0.
The same applies for the ```gradleExtractorDependencyJCenterVersion``` const.

eyalbe4 (changes requested):

PR #131, opened 2021-05-27
Fix: 'jfrog rt go get' 

gailazar300:
-----


artifactory/commands/golang/go.go
```diff
@@ -214,8 +215,36 @@ func getPackageFilePathFromArtifactory(packageName, rtTargetRepo string, authArt
...
+	return path, nil
+
+}
 
+// getFileSystemPackagePath returns a string that represents the package files cache path.
+// In some cases when the path isn't represented by the package name, instead the name represents a specific project's directory's path.
+// In this case we will scan the path until we find the package directory.
+// Example : When running 'go get github.com/golang/mock/mockgen@v1.4.1'
+//			* "mockgen" is a directory inside "mock" package ("mockgen" doesn't contain "go.mod").
+//			* go download and save the whole "mock" package in cache under 'github.com/golang/mock@v1.4.1'.
```
eyalbe4, line 231: go download and save the whole "mock" package in cache under... --> "go get" downloads and saves the whole "mock" package in the local cache under...

artifactory/commands/golang/go.go
```diff
@@ -214,8 +215,36 @@ func getPackageFilePathFromArtifactory(packageName, rtTargetRepo string, authArt
...
+func getFileSystemPackagePath(packageCachePath, name, version string) (string, error) {
+	separator := clientutils.GetFileSeparator()
```
eyalbe4, line 233: We'll need to see if this searator works well on Windows. We'll need to wait and see if your tests pass on Appveyor.
I can see that we use ```filepath.Separator``` in other places in the jfrog-clie-core code base.

artifactory/commands/golang/go_test.go
```diff
@@ -21,3 +27,18 @@ func TestBuildPackageVersionRequest(t *testing.T) {
 		})
 	}
 }
+
+func TestGetPackageFilesPath(t *testing.T) {
```
eyalbe4, line 31: This is not a unit test, but an integration test. It requires "go" and also Artifactory. The jfrog-cli-core test command hoeever do not aware of Artifactory, and also do not have requirement to have "go" installed on the test machine (see the Tests section in the README).
This test should probably move jfrog-cli, and be added to go integration tests.

In additopn to the above, I think that the ```getFileSystemPackagePath``` does require a unit test. In order to unit test this function, we don't need "go" or even the file system. It should be "table tested" to see that its logic is good.

artifactory/utils/golang/project/project.go
```diff
@@ -249,8 +249,7 @@ func (project *goProject) readModFile() error {
 	modFilePath := filepath.Join(project.projectPath, "go.mod")
 	modFileExists, _ := fileutils.IsFileExists(modFilePath, true)
 	if !modFileExists {
-		log.Info("Could not find go.mod in ", project.projectPath)
-		return nil
+		return errors.New("Could not find project's go.mod in " + project.projectPath)
```
eyalbe4, line 252: This error should be wrapped with CheckError.

eyalbe4 (changes requested):

PR #134, opened 2021-06-17
Build link fix

Ohad2Levy:
-----

### DIFF
--- a/artifactory/commands/buildinfo/publish.go
+++ b/artifactory/commands/buildinfo/publish.go
@@ -103,7 +103,7 @@ func (bpc *BuildPublishCommand) Run() error {
 	timestamp := buildTime.UnixNano() / 1000000
 	artVersioin, _ := servicesManager.GetVersion()
 
-	if artVersioin[0:1] == "6" &&  {
+	if artVersioin[0:1] == "6" {
 		log.Info("Build info successfully deployed. Browse it in Artifactory under " +
 			bpc.serverDetails.GetUrl() +
 			"artifactory/webapp/builds/" +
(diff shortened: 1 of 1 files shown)


artifactory/commands/buildinfo/publish.go
```diff
@@ -91,6 +93,35 @@ func (bpc *BuildPublishCommand) Run() error {
...
 	}
+
+	buildTime, err := time.Parse(buildinfo.TimeFormat, buildInfo.Started)
+	if errorutils.CheckError(err) != nil {
+		return err
+	}
+
+	// Convert from nanoseconds to milliseconds
+	timestamp := buildTime.UnixNano() / 1000000
+	artVersioin, _ := servicesManager.GetVersion()
```
eyalbe4, line 104: Let's not ignore the err returned from the ```GetVersion()``` method.

artifactory/commands/buildinfo/publish.go
```diff
@@ -91,6 +93,35 @@ func (bpc *BuildPublishCommand) Run() error {
...
 		return err
 	}
+
+	buildTime, err := time.Parse(buildinfo.TimeFormat, buildInfo.Started)
+	if errorutils.CheckError(err) != nil {
+		return err
+	}
+
+	// Convert from nanoseconds to milliseconds
+	timestamp := buildTime.UnixNano() / 1000000
```
eyalbe4, line 103: The calculated ```timestamp``` is not needed in the case if Artifactory 6. Let's therefore redesign the code, so that it is calculated only if needed.

artifactory/commands/buildinfo/publish.go
```diff
@@ -91,6 +93,35 @@ func (bpc *BuildPublishCommand) Run() error {
...
+	artVersioin, _ := servicesManager.GetVersion()
+
+	if artVersioin[0:1] == "6" {
```
eyalbe4, line 106: I suggest casting the value of ```artVersioin[0:1]``` into an int, so that we can check if the value is equals or smaller than 6. This way, this if statement. will also catch the case of Artifactory 5, 4 and 3.

artifactory/commands/buildinfo/publish.go
```diff
@@ -91,6 +93,35 @@ func (bpc *BuildPublishCommand) Run() error {
 	if err != nil {
 		return err
 	}
+
+	buildTime, err := time.Parse(buildinfo.TimeFormat, buildInfo.Started)
+	if errorutils.CheckError(err) != nil {
+		return err
+	}
+
+	// Convert from nanoseconds to milliseconds
```
eyalbe4, line 102: Let's move all the code that builds the build-info URL into a separate private method of this struct, and then add a unit test for it (in publish_test.go). The unit test can then check that the URL is build properly  when the method receieved different versions.

eyalbe4 (changes requested):

Ohad2Levy:
This can be closed please see:
https://github.com/jfrog/jfrog-cli-core/pull/216

PR #142, opened 2021-06-27
Print scan results table

asafgabai:
-----


xray/utils/resultstable_test.go
```diff
@@ -0,0 +1,24 @@
+package scan
```
eyalbe4, line 1: This package should be changed from ```scan``` to ```utils```.

xray/utils/resultstable.go
```diff
@@ -0,0 +1,96 @@
+package scan
```
eyalbe4, line 1: This package should be changed from ```scan``` to ```utils```.
The same goes for the ```downloadindexer.go``` file.

eyalbe4 (changes requested):

PR #188, opened 2021-07-20
Fix: Get deploy target repository while using old gradle plugins versions

gailazar300:
-----
Relevant deployment repository will be written by the buildinfo project to the diplyableArtifacts file starting from version 2.24.12 of build-info-extractor-gradle.
In case of a gradle project with a configuration of 'usePlugin=true' it's possible that an old build-info-extractor-gradle version is being used.
In this case, deployment repository will be read from the local project configuration file.


artifactory/commands/utils/result.go
```diff
@@ -54,21 +67,50 @@ func UnmarshalDeployableArtifacts(filePath, rtUrl string) (*Result, error) {
...
 	return result, nil
 }
 
+func GetDeployerUrlAndRepo(modulesMap *map[string][]clientutils.DeployableArtifactDetails, configPath string) (string, string, error) {
+	repo := getTargetRepoFromMap(modulesMap)
+	vConfig, err := utils.ReadConfigFile(configPath, utils.YAML)
+	if err != nil {
+		return "", "", err
+	}
+	// Relevant deploy repository will be written by the buildinfo project in diplyableArtifacts file from gradle extractor v2.24.12.
```
eyalbe4, line 91: Relevant deploy repository... --> The relevant deployment repository...

...in diplyableArtifacts file from gradle extractor v2.24.12. --> ...to the diplyableArtifacts file starting from version 2.24.12 of build-info-extractor-gradle.

artifactory/commands/utils/result.go
```diff
@@ -54,21 +67,50 @@ func UnmarshalDeployableArtifacts(filePath, rtUrl string) (*Result, error) {
...
+	// In case of a gradle project with a configuration of 'usePugin=true' its possible that an old buildinfo version is being used.
```
eyalbe4, line 92: ...its possible... --> ...it's possible...
...buildinfo version... --> ...build-info-extractor-gradle version...

artifactory/commands/utils/result.go
```diff
@@ -54,21 +67,50 @@ func UnmarshalDeployableArtifacts(filePath, rtUrl string) (*Result, error) {
...
+	// Deploy repository will be read from the configuration file.
```
eyalbe4, line 93: Deploy repository will be read from the configuration file.
-->
In this case, the value of "repo" will be empty, and the deployment repository will be therefore read from the local project configuration file.

artifactory/commands/utils/result_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"strconv"
+	"strings"
+	"testing"
+)
+
+// Check a case that targetRepository is not written in a deployableArtifacts file and needs to be read from the config file.
+func TestUnmarshalDeployableArtifacts(t *testing.T) {
+	err, cleanUpJfrogHome := CreateTempJfrogConfig()
+	defer cleanUpJfrogHome()
+	// DeployableArtifact file is being change during runtime so a copy needs to be created.
```
eyalbe4, line 22: ...change during runtime so... --> ...changed at runtime so...

artifactory/commands/utils/result_test.go
```diff
@@ -0,0 +1,96 @@
...
+	for transferDetails := new(clientutils.FileTransferDetails); r.reader.NextRecord(transferDetails) == nil; transferDetails = new(clientutils.FileTransferDetails) {
+		assert.True(t, strings.HasPrefix(transferDetails.TargetPath, "http://localhost:8080/artifactory/"))
+		assert.True(t, strings.Contains(transferDetails.TargetPath, "gradle-local-repo"))
+	}
+}
+
+func createTempDeployableArtifactFile() (string, error) {
+	testsDataGradlePath := getTestsDataGradlePath()
+	summary, err := os.Open(path.Join(testsDataGradlePath, "deployableArtifacts", "summary"))
+	tmpSummary, err := os.Create(path.Join(testsDataGradlePath, "deployableArtifacts", "tmpSummary"))
```
eyalbe4, line 36: We're not supposed to add or modify files under the source code directory. That includes the testdata dir. We should create all the temp files under one directory. The code should use one locations (and API to get it).

In addition, jfrog-client-go already includes a ```CopyFile``` function. Let's reuse it instead of creating a new one.

artifactory/commands/utils/result_test.go
```diff
@@ -0,0 +1,96 @@
...
+		return "", err
+	}
+	return tmpSummary.Name(), nil
+}
+
+func getTestsDataGradlePath() string {
+	return path.Join("..", "testdata", "gradle")
+}
+
+func CreateTempJfrogConfig() (err error, cleanUp func()) {
```
eyalbe4, line 51: Looks like this function can be made private.
Actually, the implementation of this function seem wrong. We should avoid duplicating internal logic of the config mechanism. For example - this is something we should never do - 
```
	versionString := ".v" + strconv.Itoa(coreutils.GetConfigVersion())
	configFile, err := os.Create(filepath.Join(homedir, coreutils.JfrogConfigFile+versionString))
```
I suggest rethinking this design.

utils/config/config.go
```diff
@@ -188,7 +188,7 @@ func readConf() (*ConfigV5, error) {
 		// No config file was found, returns a new empty config.
 		return config, nil
 	}
-	content, err = convertIfNeeded(content)
+	content, err = ConvertIfNeeded(content)
```
eyalbe4, line 191: I really think ```ConvertIfNeeded``` should remain private.

eyalbe4 (changes requested):

utils/config/config.go
```diff
@@ -651,3 +652,42 @@ func (missionControlDetails *MissionControlDetails) GetAccessToken() string {
...
+	`
+	content, err := convertIfNeeded([]byte(configuration))
+	if err != nil {
+		return
+	}
+	configFilePath, err := getConfFilePath()
+	configFile, err := os.Create(configFilePath)
+	defer configFile.Close()
+	_, err = configFile.Write(content)
+	if err != nil {
```
eyalbe4, line 689: This if statement seem redundant, becuase the next line returns the error anyway.

eyalbe4 (changes requested):

utils/config/config_test.go
```diff
@@ -332,10 +314,6 @@ func verifyEncryptionStatus(t *testing.T, original, actual *ConfigV5, encryption
 	}
 }
 
-func copyResources(t *testing.T, sourcePath string, destPath string) {
-	assert.NoError(t, fileutils.CopyDir(sourcePath, destPath, true, nil))
-}
-
 func assertCertsMigration(t *testing.T) {
```
eyalbe4, line 317: Missing new-line before this function.

eyalbe4 (changes requested):

PR #193, opened 2021-07-26
Get Xray Indexer version from the Indexer itself

asafgabai:
-----

### DIFF
--- a/utils/coreutils/utils.go
+++ b/utils/coreutils/utils.go
@@ -287,6 +287,14 @@ func GetJfrogPluginsDir() (string, error) {
 	return filepath.Join(homeDir, JfrogPluginsDirName), nil
 }
 
+func GetJfrogLocksDir() (string, error) {
+	homeDir, err := GetJfrogHomeDir()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(homeDir, JfrogLocksDirName), nil
+}
+
 // Ask a yes or no question, with a default answer.
 func AskYesNo(promptPrefix string, defaultValue bool) bool {
 	defStr := "[n]"
--- a/utils/coreutils/utils.go
+++ b/utils/coreutils/utils.go
@@ -295,6 +295,15 @@ func GetJfrogLocksDir() (string, error) {
 	return filepath.Join(homeDir, JfrogLocksDirName), nil
 }
 
+func GetJfrogConfigLockDir() (string, error) {
+	configLockDirName := "config"
+	locksDirPath, err := GetJfrogLocksDir()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(locksDirPath, configLockDirName), nil
+}
+
 // Ask a yes or no question, with a default answer.
 func AskYesNo(promptPrefix string, defaultValue bool) bool {
 	defStr := "[n]"
--- a/xray/utils/downloadindexer.go
+++ b/xray/utils/downloadindexer.go
@@ -5,9 +5,11 @@ import (
 	"fmt"
 	"github.com/jfrog/gofrog/io"
 	"github.com/jfrog/jfrog-cli-core/v2/utils/coreutils"
+	"github.com/jfrog/jfrog-cli-core/v2/utils/lock"
 	"io/ioutil"
 	"net/http"
 	"os"
+	"path"
 	"path/filepath"
 	"runtime"
 	"sort"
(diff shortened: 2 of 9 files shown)


xray/utils/downloadindexer.go
```diff
@@ -49,33 +55,123 @@ func DownloadIndexerIfNeeded(xrayManager *xray.XrayServicesManager) (string, err
 	}
```
eyalbe4, line 56: The following lines in this function:
```
	if err != nil {
		return "", err
	}
	if exists {
		return indexerPath, nil
	}
```
can be shortened to:
```
	if exists || err != nil {
		return indexerPath, err
	}
```

xray/utils/downloadindexer.go
```diff
@@ -49,33 +55,123 @@ func DownloadIndexerIfNeeded(xrayManager *xray.XrayServicesManager) (string, err
...
+	indexerPath := filepath.Join(tempDirPath, indexerFileName)
+	err = os.Chmod(indexerPath, 0777)
+
+	indexerVersion, err := getIndexerVersion(indexerPath)
+	if err != nil {
+		return "", err
+	}
+	newDirPath := filepath.Join(indexerDirPath, indexerVersion)
+
+	// In case of a hot upgrade of Xray in progress, the version of the downloaded indexer might be different of the Xray version we got above,
```
eyalbe4, line 110: ...different of the Xray... --> ...different from the Xray...

xray/utils/downloadindexer.go
```diff
@@ -49,33 +55,123 @@ func DownloadIndexerIfNeeded(xrayManager *xray.XrayServicesManager) (string, err
...
+	// so the indexer we just downloaded may already exist.
+	newDirExists, err := fileutils.IsDirExists(newDirPath, false)
+	if err != nil {
+		return "", err
+	}
+	if newDirExists {
+		err = os.RemoveAll(tempDirPath)
+	} else {
+		err = os.Rename(tempDirPath, newDirPath)
```
eyalbe4, line 119: We're not allowed to use ```os.Rename```, because it will fail when running inside a docker image. Instead, you should use ```fileutils.MoveFile```.

xray/utils/downloadindexer.go
```diff
@@ -21,6 +25,8 @@ import (
 const (
 	indexerFileName     = "indexer-app"
 	graphScanMinVersion = "3.29.0"
+	indexerDirName      = "xray-indexer"
+	tempIndexerDirName  = "temp"
 )
 
 func DownloadIndexerIfNeeded(xrayManager *xray.XrayServicesManager) (string, error) {
```
eyalbe4, line 32: If multiple jfrog-cli processes are running on the same machine, some of them may attempt to download the indexer at rthe same time, which can lead to issues.
I therefore suggest putting in place a cross-process lock on this function. We've already implemented such a locking mechanism on the config, which has been serving us well for a long time , so you can reuse this locking mechism.

xray/utils/downloadindexer.go
```diff
@@ -49,33 +55,123 @@ func DownloadIndexerIfNeeded(xrayManager *xray.XrayServicesManager) (string, err
...
+}
+
+func deleteOldIndexers(indexerDirPath string) error {
+	indexerDirExists, err := fileutils.IsDirExists(indexerDirPath, false)
+	if err != nil {
+		return err
+	}
+	if !indexerDirExists {
+		return nil
+	}
```
eyalbe4, line 147: The following code:
```
	if err != nil {
		return err
	}
	if !indexerDirExists {
		return nil
	}
```
can be simplified to:
```
	if !indexerDirExists || err != nil {
		return err
	}
```

xray/utils/downloadindexer_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/stretchr/testify/assert"
+	"os"
+	"path"
+	"testing"
+)
+
+func TestDeleteOldIndexers(t *testing.T) {
+	testsDir, err := fileutils.CreateTempDir()
+	assert.NoError(t, err)
+	defer fileutils.RemoveTempDir(testsDir)
```
eyalbe4, line 14: This defer should be replaced with:
```
	defer func() {
		assert.NoError(t, fileutils.RemoveTempDir(testsDir))
	}()
```

xray/utils/downloadindexer_test.go
```diff
@@ -0,0 +1,56 @@
...
+	assert.True(t, checkIndexerExists(t, indexersDirsPaths[2]))
+}
+
+func createDummyIndexer(t *testing.T, dirPath string) {
+	err := os.MkdirAll(dirPath, 0777)
+	assert.NoError(t, err)
+	fullPath := path.Join(dirPath, indexerFileName)
+	file, err := os.Create(fullPath)
+	assert.NoError(t, err)
+	defer file.Close()
```
eyalbe4, line 46: This defer line should be replaced with
```
	defer func() {
		assert.NoError(t, file.Close())
	}()
```

eyalbe4 (changes requested):

utils/coreutils/utils.go
```diff
@@ -287,6 +287,14 @@ func GetJfrogPluginsDir() (string, error) {
 	return filepath.Join(homeDir, JfrogPluginsDirName), nil
 }
 
+func GetJfrogLocksDir() (string, error) {
```
eyalbe4, line 290: I suggest adding a new ```GetJfrogConfigLocksDir``` function (which uses the ```GetJfrogLocksDir``` function) right after this function. This will allow us to remove the "join" from the code which creates the config lock, and also remove the ```config.ConfigLockDirName``` const.

xray/utils/downloadindexer.go
```diff
@@ -21,6 +26,8 @@ import (
 const (
 	indexerFileName     = "indexer-app"
 	GraphScanMinVersion = "3.29.0"
```
eyalbe4, line 28: Looks like this const can remain private (unless it is used by the jfrog-cli project).

eyalbe4 (changes requested):

xray/utils/downloadindexer.go
asafgabai: This change was merged just a few days ago (#195). It's not used by jfrog-cli yet, but it's going to be used there soon.

PR #197, opened 2021-07-29
Handle npm version prefix

RobiNino:
-----

Cannot rely on `--json` output of `npm pack` because of [npm/cli#3405](https://github.com/npm/cli/issues/3405)


utils/npm/version.go
```diff
@@ -0,0 +1,66 @@
...
+	"os/exec"
+
+	gofrogcmd "github.com/jfrog/gofrog/io"
+	"github.com/jfrog/jfrog-client-go/utils/errorutils"
+)
+
+func Version(executablePath string) (*version.Version, error) {
+	pipeReader, pipeWriter := io.Pipe()
+	defer pipeReader.Close()
+	defer pipeWriter.Close()
```
eyalbe4, line 18: The error returned from the above two Close() methods is ignored.
Here's how we can avoid ignoring this error, which inside a defer clause:
1. Change the function declaration to using named returned values - 
```
func Version(executablePath string) (version *version.Version, err error) {
```
2. Change the defer, so thay the error returned from ```pipeReader.Close()``` is returned, if the function did not return an error - 
```
	defer func() {
		e := pipeReader.Close()
		if err == nil {
			err = e
		}
		e = pipeWriter.Close()
		if err == nil {
			err = e
		}
	}()
```
Let's go over the code base and make this change in other places as well.

utils/npm/version.go
```diff
@@ -0,0 +1,66 @@
...
+	var npmError error
+
+	configListCmdConfig := createVersionCmdConfig(executablePath, pipeWriter)
+	go func() {
+		npmError = gofrogcmd.RunCmd(configListCmdConfig)
```
eyalbe4, line 23: I suggest we avoid using ```gofrogcmd.RunCmd``` when not necessary, but use go's default cmd API instead. The ```gofrog``` API is useful when the output needs to be parsed using a regexp, but as you may remember, it has its limitation. See other places the code where we use the default go API.

utils/npm/version.go
```diff
@@ -0,0 +1,66 @@
...
+
+func GetNpmVersionAndExecPath() (*version.Version, string, error) {
+	log.Debug("Getting npm executable path and version")
+	npmExecPath, err := exec.LookPath("npm")
+	if err != nil {
+		return nil, "", errorutils.CheckError(err)
+	}
+
+	if npmExecPath == "" {
+		return nil, "", errorutils.CheckError(errors.New("could not find 'npm' executable"))
```
eyalbe4, line 55: could not find 'npm' executable --> could not find 'npm' executable in PATH

eyalbe4 (changes requested):

PR #201, opened 2021-08-04
Improve ssh usability

yahavi:
-----

* Currently SSH connection is only allowed on Artifactory. When the user provides SSH platform URL, we should populate Artifactory's URL with this SSH URL.
* Username, password, and access tokens should not be provided when SSH URL is provided.
* SSH passphrase should be displayed (as `***`) when running `jfrog c show`.
* Upgrade crypto to the latest


utils/config/config.go
```diff
@@ -508,7 +508,7 @@ type ServerDetails struct {
 	User                 string `json:"user,omitempty"`
 	Password             string `json:"password,omitempty"`
 	SshKeyPath           string `json:"sshKeyPath,omitempty"`
-	SshPassphrase        string `json:"SshPassphrase,omitempty"`
+	SshPassphrase        string `json:"sshPassphrase,omitempty"`
```
eyalbe4, line 511: Dopes this require config upgrade to v6?

eyalbe4 (changes requested):

utils/config/config.go
yahavi: No. It's case-insensitive. See https://pkg.go.dev/encoding/json#Unmarshal:

> To unmarshal JSON into a struct, Unmarshal matches incoming object keys to the keys used by Marshal (either the struct field name or its tag), preferring an exact match but also accepting a case-insensitive match

PR #203, opened 2021-08-04
v1: Improve ssh usability

yahavi:
-----
    
* Currently SSH connection is only allowed on Artifactory. When the user provides SSH platform URL, we should populate Artifactory's URL with this SSH URL.
* Username, password, and access tokens should not be provided when SSH URL is provided.
* SSH passphrase should be displayed (as `***`) when running `jfrog c show`.
* Upgrade crypto to the latest

### DIFF
--- a/common/commands/config.go
+++ b/common/commands/config.go
@@ -8,7 +8,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"syscall"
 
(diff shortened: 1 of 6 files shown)


eyalbe4 (changes requested):
See my comment on https://github.com/jfrog/jfrog-cli-core/pull/201

PR #212, opened 2021-08-12
Make "M2_HOME" env variable optional

gailazar300:
-----
 Use "mvn --version" command to find maven home if not provided.

### DIFF
--- a/utils/mvn/utils.go
+++ b/utils/mvn/utils.go
@@ -60,7 +60,7 @@ func getMavenHome() (string, error) {
 		// First we will try lo look for 'mvn' in PATH.
 		mvnPath, err := exec.LookPath("mvn")
 		if err != nil || mvnPath == "" {
-			return "", errorutils.CheckError(errors.New(err.String() + "Hint: The mvn command may not be included in the PATH. Either add it to the path, or set the M2_HOME environment variable value to the maven installation directory, which is the directory which includes the bin and lib directories."))
+			return "", errorutils.CheckError(errors.New(err.Error() + "Hint: The mvn command may not be included in the PATH. Either add it to the path, or set the M2_HOME environment variable value to the maven installation directory, which is the directory which includes the bin and lib directories."))
 		}
 		log.Debug(MavenHome, " is not defined. Retrieving Maven home using 'mvn --version' command.")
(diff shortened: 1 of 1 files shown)


utils/mvn/utils.go
```diff
@@ -47,13 +48,33 @@ func RunMvn(configPath, deployableArtifactsFile string, buildConf *utils.BuildCo
...
+		output := strings.Split(strings.TrimSpace(stdout.String()), "\n")
+		// Finding the relevant "Maven home" line in command response.
+		for _, line := range output {
+			if strings.HasPrefix(line, "Maven home:") {
+				mavenHome = strings.Split(line, " ")[2]
+				break
+			}
+		}
 	}
-	return nil
```
eyalbe4, line 56: I think we should check here if mavenHome is still empty, and in this case, return an error sayong that we failed to fetch the maven home from the 'mvn --version' output. Worth adding the actual command output to the error message.

eyalbe4 (changes requested):

utils/mvn/utils.go
```diff
@@ -47,13 +48,42 @@ func RunMvn(configPath, deployableArtifactsFile string, buildConf *utils.BuildCo
...
+		output := strings.Split(strings.TrimSpace(stdout.String()), "\n")
+		// Finding the relevant "Maven home" line in command response.
+		for _, line := range output {
+			if strings.HasPrefix(line, "Maven home:") {
+				mavenHome = strings.Split(line, " ")[2]
+				break
+			}
+		}
+		if mavenHome == "" {
+			return "", errorutils.CheckError(errors.New("Can't find maven home directory. Set 'M2_HOME' environment variable or install maven on your machine.\n\"mvn --version\" command result:\n" + stdout.String()))
```
eyalbe4, line 82: Can't find maven home directory. Set 'M2_HOME' environment variable or install maven on your machine.
"mvn --version" command result:
-->
Could not find the location of the maven home directory, by running 'mvn --version' command. The command output is XXX. You also have the option of setting the M2_HOME environment variable value to the maven installation directory, which is the directory which includes the bin and lib directories.

eyalbe4 (changes requested):

PR #215, opened 2021-08-18
Fix Xray scan on Windows os

talarian1:
-----

### DIFF
--- a/xray/utils/downloadindexer.go
+++ b/xray/utils/downloadindexer.go
@@ -3,9 +3,6 @@ package utils
 import (
 	"errors"
 	"fmt"
-	"github.com/jfrog/gofrog/io"
-	"github.com/jfrog/jfrog-cli-core/v2/utils/coreutils"
-	"github.com/jfrog/jfrog-cli-core/v2/utils/lock"
 	"io/ioutil"
 	"net/http"
 	"os"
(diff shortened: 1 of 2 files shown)


xray/utils/downloadindexer.go
```diff
@@ -30,6 +31,8 @@ const (
 	tempIndexerDirName  = "temp"
 )
 
+var indexerExtension = ""
```
asafgabai, line 34: Instead of keeping only the extension as a variable here, I suggest that this variable keeps the Indexer file's name (in Windows, it will include the .exe suffix).
This way, you'll be able to avoid concatenating the extension to the file name every time you need it.

asafgabai (changes requested):

xray/utils/downloadindexer.go
```diff
@@ -178,3 +182,12 @@ func deleteOldIndexers(indexerDirPath string) error {
 
 	return nil
 }
+
+func getIndexerExtension(os string) string {
```
eyalbe4, line 186: It looks like this functions does not need to receive the OS as an argument. It can figure it out by itself.
Also, I'mm not sure separating the indexer app file name from its extension is the simplest approach. I suggest we simply rename this function to ```getIndexerBinaryname``` and have it return either "indexer-app" or "indexer-app.exe". This allows us to remove both the ```indexerExtension``` variable and also the ```indexerFileName``` constants. Anyone who'll need the idexer app name will simply call this function.

eyalbe4 (changes requested):

xray/utils/downloadindexer.go
eyalbe4: Now I see that this is also what @asafgabai suggested above :)

PR #243, opened 2021-10-25
Add Pipenv support

sverdlov93:
-----

Pipenv support for pipenv-config / pipenv-install / audit-pipenv commands


artifactory/utils/python/python.go
```diff
@@ -0,0 +1,66 @@
...
+	if username != "" && password != "" {
+		rtUrl.User = url.UserPassword(username, password)
+	}
+	rtUrl.Path += "api/pypi/" + repository + "/simple"
+
+	pc.CommandArgs = append(pc.CommandArgs, rtUrl.String())
+	return nil
+}
+
+type PythonCmd struct {
```
eyalbe4, line 59: Looks like this struct (and its methods) can be moved out of this file, and put in the same file as ```PythonCommand```. I think this will also allow us to make the ```PythonCmd``` struct private.

utils/python/pipdeptree/scriptCreator.go
```diff
@@ -42,7 +42,7 @@ func main() {
 	// Replace all backticks ( ` ) with a single quote ( ' )
 	pyFileString := strings.ReplaceAll(string(pyFile), "`", "'")
 	// Add it the relevant package
```
eyalbe4, line 44: This comment is unclear.

utils/python/pipenvutils.go
```diff
@@ -0,0 +1,83 @@
...
+	packages, err := runPipenvGraph(venvDir)
+	if err != nil {
+		return nil, err
+	}
+
+	// Parse the result.
+	return parseDependenciesToList(packages)
+}
+
+// Executes the pipenv install and pipenv graph
```
eyalbe4, line 38: // Executes the pipenv install and pipenv graph
-->
// Executes pipenv install and pipenv graph.

utils/python/pipenvutils.go
```diff
@@ -0,0 +1,83 @@
...
+// Returns a dependency map of all the installed pip packages in the current environment to and another list of the top level dependencies
+func GetPipenvDependenciesGraph(venvDir string) (map[string][]string, []string, error) {
+	// run pipenv install
```
eyalbe4, line 41: run --> Run

utils/python/pipenvutils.go
```diff
@@ -0,0 +1,83 @@
...
+	packages, err := runPipenvGraph(venvDir)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Parse the result.
+	return parseDependenciesToGraph(packages)
+}
+
+// Executes the pipenv graph and returns a dependency map of all the installed pip packages in the current environment to and another list of the top level dependencies
```
eyalbe4, line 57: This sentence is too long, and its last part - "...to and another list of the top level dependencies" is unclear and seems detached from the beginning of the senstence.

utils/python/pipenvutils.go
```diff
@@ -0,0 +1,83 @@
...
+	data, err := runPythonCommand("pipenv", []string{"graph", "--json"}, getPipenvEnvironmentString(venvDir))
+	if err != nil {
+		return nil, err
+	}
+	// Parse into array.
+	packages := make([]pythonDependencyPackage, 0)
+	if err := json.Unmarshal(data, &packages); err != nil {
+		return nil, errorutils.CheckError(err)
+	}
+	// Parse the result.
```
eyalbe4, line 68: Seems an incorrect description.

utils/python/pipenvutils.go
```diff
@@ -0,0 +1,83 @@
...
+	return packages, nil
+}
+
+// Parse pip-dependency-map raw output to dependencies map (mapping dependency to his child deps) and top level deps list
```
eyalbe4, line 72: Let's rephrase. By reading the current description, it is pretty hard to understand what;s beibg parsed. Specifically - "... and top level deps list" is unclear.

utils/python/utils.go
```diff
@@ -0,0 +1,70 @@
+package python
+
```
eyalbe4, line 2: It looks nlike the code in this file can be merged with the code in ```utils/python/piputils.go ```. Not sure if there's a need for this split.
Let's also merge the ```populatePythonDependencyTree``` utility function into this single util file.

artifactory/commands/python/python.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	if !validPath {
+		log.Debug("Could not find setup.py file in current directory:", wd)
+		return "", nil
+	}
+
+	return filePath, nil
+}
+
+func (pc *PythonCommand) cleanBuildInfoDir() {
```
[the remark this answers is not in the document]
sverdlov93: we use cleanBuildInfoDir only after we catch an error so there is already a previous error to throw

eyalbe4 (changes requested):

eyalbe4 (changes requested):
In the below method
```
func (bsc *BuildScanCommand) runBuildScanAndPrintResults(xrayManager *xray.XrayServicesManager, params services.XrayBuildParams) (bool, error) {
	buildScanResults, err := xrayManager.BuildScan(params)
	if err != nil {
		return false, err
	}
	scanResponseArray := []services.ScanResponse{{Violations: buildScanResults.Violations, XrayDataUrl: buildScanResults.MoreDetailsUrl}}
	log.Info("Xray data is available to view at: " + buildScanResults.MoreDetailsUrl)
	err = xrutils.PrintScanResults(scanResponseArray, bsc.outputFormat == xrutils.Table, false, false, false)
	if err != nil {
		return false, err
	}
	return xrutils.CheckIfFailBuild(scanResponseArray), nil
}
```

Xray data is available to view at:
-->
Xray data is available at:

eyalbe4 (changes requested):
In the ```(bsc *BuildScanCommand) Run()``` method, let's have all comments start with a capital letter, according to our conversion.

artifactory/commands/python/python.go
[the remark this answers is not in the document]
eyalbe4: Yes - I can see that. However, this function can be used in the future by a developer from a different flow. The developer shouldn't be concerned about error being swallowed.
The right approach in my opnion, is to return the error from the function, and concatenate its message to the message of the error in the calling function.

eyalbe4 (changes requested):

PR #248, opened 2021-11-10
Add vcs-fallback feature for Go

yahavi:
-----

Depends on https://github.com/jfrog/gocmd/pulls

If a package does not exist in Artifactory, download it from VCS.
Add the `--vcs-fallback` boolean flag with default true.


artifactory/commands/golang/go.go
```diff
@@ -106,9 +109,26 @@ func (gc *GoCommand) Run() error {
 	if err != nil {
 		return err
 	}
+
+	// Extract vcs fallback from the args.
+	gc.goArg, err = gc.extractVcsFallbackFromArgs()
```
eyalbe4, line 114: Let's not ignore the error returned here.

utils/coreutils/cmdutils.go
```diff
@@ -120,6 +120,22 @@ func FindBooleanFlag(flagName string, args []string) (flagIndex int, flagValue b
...
 }
 
+// Boolean flag can be provided in one of the following forms:
+// 1. --flag=value, where value can be true/false
+// 2. --flag, here the value is true
+// Return values:
+// flagIndex - index of flagName in args.
+// flagValue - value of flagName.
+// err - error if flag exists, but we failed to extract its value.
+// If flag does not exist flagIndex = -1 with true value and nil error.
```
eyalbe4, line 130: I think that the last line in the above description isn't completely clear. The same goes for the last line in the description of the ```FindBooleanFlag```.
Actually, I suggest that instead of duplicating most of the description for this function, we'll described it as - "The same as the FindBooleanFlag function, but ...".

eyalbe4 (changes requested):

PR #250, opened 2021-11-17
Small fixes

asafgabai:
-----
Small fixes:
* Avoid closing the zip temp file twice in go-publish.
* Check if running in terminal also when printing a message in a frame.

### DIFF
--- a/artifactory/commands/golang/publish.go
+++ b/artifactory/commands/golang/publish.go
@@ -179,16 +179,15 @@ func archive(moduleName, version, projectPath, tempDir string) (name string, zip
 	if err != nil {
 		return "", nil, errorutils.CheckError(err)
 	}
+	openedFile = true
 	defer func() {
-		if !openedFile {
-			return
-		}
-		e := tempFile.Close()
-		if err == nil {
-			err = errorutils.CheckError(e)
+		if openedFile {
+			e := tempFile.Close()
+			if err == nil {
+				err = errorutils.CheckError(e)
+			}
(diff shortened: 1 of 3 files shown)


artifactory/commands/golang/publish.go
```diff
@@ -174,16 +174,21 @@ func readModFile(version, projectPath string, createArtifact bool) ([]byte, *bui
...
 	}
 	defer func() {
+		if !openedFile {
+			return
+		}
 		e := tempFile.Close()
 		if err == nil {
 			err = errorutils.CheckError(e)
 		}
 	}()
```
eyalbe4, line 190: We can simplify the following code a bit - 
```
	defer func() {
		if !openedFile {
			return
		}
		e := tempFile.Close()
		if err == nil {
			err = errorutils.CheckError(e)
		}
	}()
```
to be - 
```
	defer func() {
		if openedFile {
                   e := tempFile.Close()
                   if err == nil {
                       err = errorutils.CheckError(e)
                   }
		}
	}()
```

artifactory/commands/golang/publish.go
```diff
@@ -174,16 +174,21 @@ func readModFile(version, projectPath string, createArtifact bool) ([]byte, *bui
...
+	openedFile = true
```
eyalbe4, line 191: Loiks like this line is misplaced and should be put before the defer above.

eyalbe4 (changes requested):

Issue #264, opened 2021-11-28
Maven version validation fails when ANSI control sequences are present in the version output


zsrv:
**Describe the bug**

jfrog-cli-core v2.5.1 introduced [validation](https://github.com/jfrog/jfrog-cli-core/blob/93725a913b0ac19dd8164f9645715366636c0dee/utils/mvn/utils.go#L53) of the version of Maven that is in use, which is being done by parsing the output of `mvn --version`. This validation fails if the version line, which is found by searching for a line in the output that starts with "Apache Maven", contains extra characters at the beginning of the line.

The version of Maven that I am using prints the Maven version line in bold, with ANSI control characters at the beginning (and end) of the line.

There is also a related, but secondary, issue where "minSupportedMvnVersion" is not printed correctly (see below).

Example:

```
user@localhost:~> jfrog config add --interactive=false --url=http://example.com --access-token=x
user@localhost:~> jfrog mvn-config --repo-resolve-releases=x --repo-resolve-snapshots=x --repo-deploy-releases=x --repo-deploy-snapshots=x
[Info] maven build config successfully created.
user@localhost:~> jfrog mvn clean
[Info] Running Mvn...
[Info] Could not get maven version, by running 'mvn --version' command. JFrog CLI mvn commands requires Maven version "+minSupportedMvnVersion+" or higher.
[Error] JFrog CLI mvn commands requires Maven version "+minSupportedMvnVersion+" or higher. The Current version is: 
user@localhost:~> mvn --version
Apache Maven 3.6.3 (SUSE 3.6.3-4.2.1)
Maven home: /usr/share/maven
Java version: 11.0.13, vendor: Oracle Corporation, runtime: /usr/lib64/jvm/java-11-openjdk-11
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "5.3.18-59.34-default", arch: "amd64", family: "unix"
```

`mvn --version` output with control characters visible:

```
^[[1mApache Maven 3.6.3 (SUSE 3.6.3-4.2.1)^[[m
Maven home: /usr/share/maven
Java version: 11.0.13, vendor: Oracle Corporation, runtime: /usr/lib64/jvm/java-11-openjdk-11
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "5.3.18-59.34-default", arch: "amd64", family: "unix"
```

My workaround was to downgrade to JFrog CLI 2.6.1 (which uses jfrog-cli-core 2.5.0).

[...]

PR #265, opened 2021-11-28
  Fix maven version validation parsing

sverdlov93:
-----
Fix maven version validation parsing issue.
Following #264 issue

### DIFF
--- a/utils/mvn/utils.go
+++ b/utils/mvn/utils.go
@@ -123,7 +123,7 @@ func parseMvnHome(line string) (string, error) {
 	// Line example: 'Maven home: /usr/share/maven'
 	
 	// Remove all prefix before 'Maven' (if exists)
-	line = line[strings.Index(line,"Maven"):]
+	line = line[strings.Index(line, "Maven"):]
 	// Get version string 
 	mavenHome := strings.Split(line, " ")[2]
 	if coreutils.IsWindows() {
--- a/utils/mvn/utils.go
+++ b/utils/mvn/utils.go
@@ -8,7 +8,6 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
-	"regexp"
 	"strings"
 
 	"github.com/jfrog/jfrog-cli-core/v2/utils/coreutils"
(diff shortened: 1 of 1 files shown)


utils/mvn/utils.go
```diff
@@ -70,8 +71,9 @@ func getMavenHomeAndValidateVersion() (string, error) {
 				return "", err
 			}
 		}
```
eyalbe4, line 73: Looking at the following code - 
```
	// Finding the relevant "Maven home" line in command response.
	for _, line := range output {
		if mvnHome == "" && strings.HasPrefix(line, "Maven home:") {
			// The M2_HOME environment variable is not defined.
			// Since Maven installation can be located in different locations,
			// Depending on the installation type and the OS (for example: For Mac with brew install: /usr/local/Cellar/maven/{version}/libexec or Ubuntu with debian: /usr/share/maven),
			mvnHome, err = parseMvnHome(line)
			if err != nil {
				return "", err
			}
		}
		if strings.Contains(line, "Apache Maven") {
			versionRegex := regexp.MustCompile("Apache\\sMaven\\s(\\w[\\w-\\.]+)\\s")
			mvnVersion = versionRegex.FindStringSubmatch(line)[1]
		}
	}
```
shouldn't we break out of the loop as soon as ```mvnHome``` is not empty, right after - 
```
			mvnHome, err = parseMvnHome(line)
			if err != nil {
				return "", err
			}
```
And fo so, there's no need to check the value of mvnHome in the following if statement.
```
if mvnHome == "" && strings.HasPrefix(line, "Maven home:") {
```

In addition, the following line at the beginning of the function may set a value into the ```mvnHome``` variable - 
```
mvnHome := os.Getenv(mavenHome)
```
If mvnHome include a value, we probably shouldn't get into the loop to try and set it.

eyalbe4 (changes requested):

utils/mvn/utils.go
```diff
@@ -113,6 +120,11 @@ func runMvnVersionCommand(mavenHome string) ([]string, error) {
 }
 
 func parseMvnHome(line string) (string, error) {
+	// Line example: 'Maven home: /usr/share/maven'
```
yahavi, line 123: I think it would be cleaner to put this comment above the function with a short explanation, as Godoc.

utils/mvn/utils.go
```diff
@@ -113,6 +120,11 @@ func runMvnVersionCommand(mavenHome string) ([]string, error) {
...
+	
+	// Remove all prefix before 'Maven' (if exists)
+	line = line[strings.Index(line,"Maven"):]
```
yahavi, line 126: Please run `go fmt ./...`
```suggestion
	line = line[strings.Index(line, "Maven"):]
```

yahavi (changes requested):

PR #298, opened 2022-01-06
Add terraform command

gailazar300:
-----
Add terraform commands:
*terraform-config
*terraform-publish: find terraform modules, pack in a zip, and upload to artifactory.


artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,304 @@
...
+	moduleName string
+	provider   string
+	tag        string
+	exclusions []string
+}
+
+type TerraformPublishCommand struct {
+	TerraformCommand
+	*TerraformPublishCommandArgs
+	commandName string
```
eyalbe4, line 36: There's only one command name, so why do we need a variable for storing it?
This would have made sense if we multiple commands that used this struct.

artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,304 @@
...
+	return tpc
+}
+
+func (tpc *TerraformPublishCommand) SetExclusions(exclusions []string) *TerraformPublishCommand {
+	tpc.exclusions = exclusions
+	return tpc
+}
+
+func (tpc *TerraformPublishCommand) Run() error {
+	log.Info("Running Terraform Publish")
```
eyalbe4, line 91: Publish --> publish

artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,304 @@
...
+	success, failed, err := tpc.TerraformPublish()
+	if err != nil {
+		return err
+	}
+	tpc.result.SetSuccessCount(success)
+	tpc.result.SetFailCount(failed)
+	return nil
+}
+
+func ExtractTerraformPublishOptionsFromArgs(args []string) (namespace, provider, tag string, exclusions []string, err error) {
```
eyalbe4, line 134: Looks like this function can be made private. 
In addition, if you made this function a member of the struct, you wouldn't need to pass any arguments to it.
Also, I suggest adding a unit test for this function.

artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,304 @@
...
+	if err != nil {
+		return
+	}
+	coreutils.RemoveFlagFromCommand(&args, flagIndex, valueIndex)
+	// Extract exclusions information from the args.
+	flagIndex, valueIndex, exclusionsString, err := coreutils.FindFlag("--exclusions", args)
+	for _, singleValue := range strings.Split(exclusionsString, ";") {
+		exclusions = append(exclusions, singleValue)
+	}
+	if err != nil {
```
eyalbe4, line 158: This error check should be moved and be eight after the function which returned the error.

artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,304 @@
...
+		return
+	}
+	coreutils.RemoveFlagFromCommand(&args, flagIndex, valueIndex)
+	if len(args) != 0 {
+		err = errorutils.CheckErrorf("Unknown flag:\"" + strings.Split(args[0], "=")[0] + "\". for a terraform publish command please provide --namespace, --provider and --tag.")
```
eyalbe4, line 163: for a terraform publish command please provide --namespace, --provider and --tag
-->
For a terraform publish command, please provide --namespace, --provider, --tag and optionally --exclusions

artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+	return
+}
+
+func (ts *TerraformPublishCommand) TerraformPublish() (int, int, error) {
```
eyalbe4, line 168: Looks like this function can be made private.

artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,304 @@
...
+	go func() {
+		defer producer.Done()
+		toArchive := make(map[string]*services.ArchiveUploadData)
+		pwd, err := os.Getwd()
+		if err != nil {
+			log.Error(err)
+			errorsQueue.AddError(err)
+		}
+		// Walk and upload directories which contain '.tf' files.
+		err = filepath.WalkDir(pwd, func(path string, info fs.DirEntry, err error) error {
```
eyalbe4, line 193: I suggest having a unit test, that checks that functionality of collecting directories with .tf files. To allow for this test, we may need to slightly refactor and isolate this walk operation. 

Also, if I understand the code correctly, the code works as follows -

1. Look for directory for upload
2. Stop
2. Upload the directory
3. Continue looking for the next directory

This is really inefficient, because of the "Stop" on step #2 above. 
I may have missed something while reading the code, so please let me know if I did. I just see the "walk" and the "upload" happening inside the same go routine.

artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,304 @@
...
+
+	log.Debug("Uploaded", strconv.Itoa(totalUploaded), "artifacts.")
+	totalFailed = totalUploadAttempted - totalUploaded
+	if totalFailed > 0 {
+		log.Error("Failed uploading", strconv.Itoa(totalFailed), "artifacts.")
+	}
+	return
+}
+
+func (tp *TerraformPublishCommand) uploadParamsForTerraformPublish(moduleName, dirPath string) (*services.UploadParams, error) {
```
eyalbe4, line 264: This function never returns an error.

artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,304 @@
...
+	uploadParams.Recursive = true
+	uploadParams.CommonParams.TargetProps = specutils.NewProperties()
+	uploadParams.CommonParams.Exclusions = append(tp.exclusions, "*.git", "*.DS_Store")
+
+	return &uploadParams, nil
+}
+
+// Module's path in terraform repository : namespace/provider/moduleName/tag.zip
+func (tp *TerraformPublishCommand) getPublishTarget(moduleName string) string {
+	return filepath.ToSlash(filepath.Join(tp.repo, tp.namespace, tp.provider, moduleName, tp.tag+".zip"))
```
eyalbe4, line 279: I think that instead of using ```filepath.Join``` and then correcting its slashes with ```filepath.ToSlash```, you could have just used ```path.Join```.
The difference between ```filepath.Join``` and ```path.Join``` is that the first one, will use backslashes foe Windows, and slashes for unix, and the second one will always use slashes, which is exactly what you need.

artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// We identify a Terraform module by the existing of a '.tf' file inside the module directory.
```
eyalbe4, line 282: existing --> existence
...a '.tf' file... --> ...a file with a ".tf" extension...

artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,304 @@
...
+// isTerraformModule search for '.tf' file inside and returns true it founds at least one.
```
eyalbe4, line 283: ....tf' file... --> ...a file with a ".tf" extension...
... inside... --> ... inside the directory sent as an argument,...

I suggest adding a unit test for this function.

artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,304 @@
...
+func checkIfTerraformModule(path string) (bool, error) {
+	dirname := path + string(filepath.Separator)
+	d, err := os.Open(dirname)
+	if err != nil {
```
eyalbe4, line 287: CheckError missing here.

artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,304 @@
...
+		return false, err
+	}
+	defer d.Close()
```
eyalbe4, line 290: The error returned here shouldn't be ignored. you should use this pattern for errors inside defer clauses - 
```
	defer func() {
		for _, reader := range temp {
			e := reader.Close()
			if err == nil {
				err = errorutils.CheckError(e)
			}
		}
	}()
``` 
This pattern is very important and should be always used when a function used with defer returns an error.

artifactory/utils/terraform.go
```diff
@@ -0,0 +1,10 @@
+package utils
+
+import (
+	"github.com/jfrog/jfrog-cli-core/v2/utils/config"
+	"github.com/jfrog/jfrog-client-go/artifactory"
+)
+
+func CreateTerraformServiceManager(serverDetails *config.ServerDetails, httpRetries, httpRetryWaitMilliSecs int, dryRun bool) (artifactory.ArtifactoryServicesManager, error) {
```
eyalbe4, line 8: This function seem to be unused, so this entire file can be probably removed.

eyalbe4 (changes requested):

artifactory/commands/terraform/terraformpublish.go
gailazar300: Don't we need a command name for the "usage report"?

artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,365 @@
...
+	return nil
+}
+
+func (tpc *TerraformPublishCommand) preparePrerequisites() error {
+	err := tpc.extractTerraformPublishOptionsFromArgs(tpc.args)
+	if err != nil {
+		return err
+	}
+	if tpc.namespace == "" || tpc.provider == "" || tpc.tag == "" {
+		return errorutils.CheckErrorf("Wrong number of arguments. for a terraform publish command please provide --namespace, --provider and --tag.")
```
eyalbe4, line 113: Wrong number of arguments. for a terraform publish command please provide --namespace, --provider and --tag.
-->
The --namespace, --provider and --tag options are mandatory.

artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+	artDetails, err := tpc.serverDetails.CreateArtAuthConfig()
+	if err != nil {
+		return err
+	}
+	return utils.ValidateRepoExists(tpc.repo, artDetails)
+}
+
+func (tpc *TerraformPublishCommand) publish() error {
+	log.Debug("Deploying terraform module.")
```
eyalbe4, line 126: Deploying terraform module.
-->
Deploying terraform module....

artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,365 @@
...
+		// Walk and upload directories which contain '.tf' files.
+		err = tpc.walkDirAndUploadTerraformModules(pwd, producer, errorsQueue, uploadSummary, addTaskWithError)
+		if err != nil && err != io.EOF {
+			log.Error(err)
+			errorsQueue.AddError(err)
+		}
+	}()
+}
+
+// ProduceTaskFunk Provided as a parameter to 'walkDirAndUploadTerraformModules' function for testing reasons.
```
eyalbe4, line 202: ProduceTaskFunk Provided as a parameter... --> ProduceTaskFunk is provided as an argument...
reasons --> purposes

artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,365 @@
...
+	return producer.AddTaskWithError(uploadService.CreateUploadAsZipFunc(uploadSummary, target, archiveData, errorsQueue), errorsQueue.AddError)
+}
+
+func (tpc *TerraformPublishCommand) walkDirAndUploadTerraformModules(pwd string, producer parallel.Runner, errorsQueue *clientutils.ErrorsQueue, uploadSummary *specutils.Result, produceTaskFunk ProduceTaskFunk) error {
+	return filepath.WalkDir(pwd, func(path string, info fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		pathInfo, e := os.Lstat(path)
+		if e != nil {
```
eyalbe4, line 215: Missing CheckError.

artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,365 @@
...
+			if archiveData.GetWriter().IsEmpty() {
+				return filepath.SkipDir
+			}
+			uploadService := createUploadServiceManager(tpc.serverDetails, errorsQueue)
+			_, e = produceTaskFunk(producer, uploadService, uploadSummary, uploadParams.Target, archiveData, errorsQueue)
+			if e != nil {
+				log.Error(e)
+				errorsQueue.AddError(e)
+			}
+			// SkipDir will not stop the walk, but will jump to the next directory.
```
eyalbe4, line 251: ...but will jump to the next directory. --> ...but it will make us jump to the next directory.

artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,365 @@
...
+		}
+		return nil
+	})
+}
+
+func createTerraformArchiveUploadData(uploadParams *services.UploadParams, errorsQueue *clientutils.ErrorsQueue) *services.ArchiveUploadData {
+	archiveData := services.ArchiveUploadData{}
+	var err error
+	archiveData.SetUploadParams(services.DeepCopyUploadParams(uploadParams))
+	//archiveData.SetUploadParams(uploadParams)
```
eyalbe4, line 262: This comment should be probably removed.

artifactory/commands/terraform/terraformpublish.go
```diff
@@ -0,0 +1,365 @@
...
+		log.Error(err)
+		errorsQueue.AddError(err)
+	}
+	uploadService := services.NewUploadService(serviceManager.Client())
+	uploadService.ArtDetails = serviceManager.GetConfig().GetServiceDetails()
+	uploadService.Threads = serviceManager.GetConfig().GetThreads()
+	return uploadService
+}
+
+func GetSaveTaskInContentWriterFunc(archiveData *services.ArchiveUploadData, uploadParams services.UploadParams, errorsQueue *clientutils.ErrorsQueue) services.UploadDataHandlerFunc {
```
eyalbe4, line 285: I think this function can be made private.

eyalbe4 (changes requested):

artifactory/commands/terraform/terraformpublish_test.go
```diff
@@ -0,0 +1,141 @@
...
+	return 0, tests.ValidateListsIdentical([]string{"a.tf"}, paths)
+}
+
+func mockProduceTaskFunk(archiveData *services.ArchiveUploadData) (paths []string, err error) {
+	archiveData.GetWriter().GetFilePath()
+	archiveDataReader := content.NewContentReader(archiveData.GetWriter().GetFilePath(), archiveData.GetWriter().GetArrayKey())
+	defer func() {
+		e := archiveDataReader.Close()
+		if e == nil {
+			err = e
```
eyalbe4, line 134: "err" is always nil here, so why do we need this if statement?

artifactory/commands/terraform/terraformpublish_test.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+	}()
+	for uploadData := new(services.UploadData); archiveDataReader.NextRecord(uploadData) == nil; uploadData = new(services.UploadData) {
+		paths = append(paths, uploadData.Artifact.TargetPathInArchive)
+	}
+	return paths, nil
```
eyalbe4, line 140: return paths, nil
-->
return

eyalbe4 (changes requested):

PR #216, opened 2021-08-18
New build link fix

Ohad2Levy:
- [x] All [tests](https://github.com/jfrog/jfrog-cli-core#tests) passed. If this feature is not already covered by the tests, I added new tests.
- [x] This pull request is on the dev branch.
- [x] I used gofmt for formatting the code before submitting the pull request.
-----

PR #323, opened 2022-01-26
V1 - Bug fix - Build-info link is wrong when used with projects

eyalbe4:
-----
This PR fixes an issue for V1 and is a duplicate of https://github.com/jfrog/jfrog-cli-core/pull/216, which has been merged to V2 a few months ago.
It fixes an issue, whereby the build-info link displayed in the log when the build-info is published, is incorrect, when the build-info is published as part of a project.
This PR depends on https://github.com/jfrog/jfrog-client-go/pull/513


artifactory/commands/buildinfo/publish.go
```diff
@@ -91,12 +96,50 @@ func (bpc *BuildPublishCommand) Run() error {
 	if err != nil {
 		return err
 	}
+
+	buildLink, err := bpc.constructBuildInfoUiUrl(servicesManager, buildInfo.Started)
```
RobiNino, line 100: I suggest returning before this construction if on `dryRun`. The preview build was already logged to the terminal by the client, and a link was not printed prior to these changes.

utils/lock/utils_unix.go
```diff
@@ -1,3 +1,4 @@
+//go:build linux || darwin || freebsd
```
RobiNino, line 1: Is this relevant to the PR?

RobiNino (changes requested):

utils/lock/utils_unix.go
eyalbe4: It is not, but this is something go 1.17 automatically modifies. It did the same for v2 as well, after we started releasing it with go 1.17.
I'll make sure that v1 is also moved to go 1.17.

artifactory/commands/buildinfo/publish.go
eyalbe4: Good point @RobiNino. I'll add this change to v2 as well.

Issue #350, opened 2022-03-07
File-Locking for ~/.jfrog/jfrog-cli.conf.v5 is unreliable


HWoidt:
**Describe the bug**
When concurrently executing configuration commands from jfrog cli, the config file in  ~/.jfrog/jfrog-cli.conf.v5 may become corrupted. This causes all jfrog cli commands that are executed later to fail with the error message "[Error] invalid character 't' after top-level value" (See corrupted json below under 'Screenshots')

**To Reproduce**
The relevant operations are:
1. https://github.com/jfrog/jfrog-cli-core/blob/a1d969576fdc05f898f04ea3f68c60f43292dab8/utils/lock/lock.go#L41
2. https://github.com/jfrog/jfrog-cli-core/blob/a1d969576fdc05f898f04ea3f68c60f43292dab8/utils/lock/lock.go#L55
3. https://github.com/jfrog/jfrog-cli-core/blob/a1d969576fdc05f898f04ea3f68c60f43292dab8/utils/lock/lock.go#L151

The files in the lockfile directory are sorted according to the timestamp embedded in the filename. The oldest file containing a PID that is still running "wins" the lock and may go ahead to change the config file.
A possible cause for the observed behavior is the following sequence of events where two processes P1 and P2 want to access the config file concurrently:
- P1: time.now()
- P2: time.now()
- P2: os.OpenFile()
- P2: fileutils.ListFiles() -> yay I'm the oldest one -> goes ahead to change the config file
- P1: os.OpenFile()
- P1: fileutils.ListFiles() -> yay I'm the oldes... ☠️

**Expected behavior**
The lockfile mechanism works reliable.

**Screenshots**
```
$ cat ~/.jfrog/jfrog-cli.conf.v5
{
  "servers": [],
  "version": "5"
}toryUrl": "*redacted*",
      "user": "*redacted*",
      "password": "*redacted*",
      "serverId": "*redacted*",
      "isDefault": true
    }
  ],
  "version": "5"
}
```

**Versions**
- JFrog CLI core version: master
- JFrog CLI version (if applicable): 2.12.1
- Artifactory version: 7.33.12

**Additional context**
We have multiple Azure-Pipeline Agents running on the same self-hosted Ubuntu 20.04.3 machine, executing tasks from the Jfrog-Azure-Devops Plugin in parallel. The plugin uses jfrog-cli internally.

**Possible Solution**
[...]

PR #352, opened 2022-03-09
Fixing config file lock mechanism

sverdlov93:
-----
Fix config commands locking mechanism.
Make all config commands that write to config file  - **atomic**. 

Also, get lock's file modified time from os, instead of using time.Now() (as suggested by @HWoidt)

should fix https://github.com/jfrog/jfrog-cli-core/issues/350


sverdlov93:
Hi @HWoidt,
First of all, Thanks for reporting this issue and for your very informative logs and data.
Our apologies for the inconvenience caused by this.

I am currently working on the fix, and have some more changes and tests to add.
I will update here as soon as it's done and ready for merge.

sverdlov93:
Hi, @HWoidt.
The PR is waiting for review but if you want to test it on your machine you can pull [JFrog CLI PR]( https://github.com/sverdlov93/jfrog-cli/tree/config-concurrency-test) and run `./build/build.sh` to get the executable.

common/commands/config.go
```diff
@@ -2,26 +2,34 @@ package commands
...
 	"github.com/jfrog/jfrog-cli-core/v2/artifactory/utils"
 	"github.com/jfrog/jfrog-cli-core/v2/utils/config"
-	"github.com/jfrog/jfrog-cli-core/v2/utils/lock"
 	clientutils "github.com/jfrog/jfrog-client-go/utils"
 	"github.com/jfrog/jfrog-client-go/utils/errorutils"
 	"github.com/jfrog/jfrog-client-go/utils/io/fileutils"
 	"github.com/jfrog/jfrog-client-go/utils/log"
 )
 
+type CmdType string
```
eyalbe4, line 24: CmdType --> ConfigAction

common/commands/config.go
```diff
@@ -34,10 +42,11 @@ type ConfigCommand struct {
 	serverId         string
 	// For unit tests
 	disablePromptUrls bool
+	cmdType           CmdType
```
eyalbe4, line 45: cmdType --> configAction

eyalbe4 (changes requested):

HWoidt (commented):
Hi @sverdlov93 ,
Sorry for letting you wait. I had other stuff on my plate.
I have now tested your changes on our Ubuntu 20.04 machine with the bash script I provided earlier:
1. When running tasks concurrently the command often fails because the Stat-call (lock.go:169) on an already deleted lock-file causes the command to fail instead of being retried. I changed this locally and made the following observations with the Stat error ignored.
2. There sometimes is a deadlock, both processes are waiting until their lock-retry loop runs out after ~2 minutes and then return with a failure. I suspect this to be due to two files having the same timestamp. This tie can occur with higher frequency than before because the filesystem timestamps have [lower resolution](https://stackoverflow.com/questions/14392975/timestamp-accuracy-on-ext4-sub-millsecond).
3. The corruption can still be observed, although with vastly lower probability. This occurred maybe 2 times in the 3h that I left it running with 40 concurrent processes. I fear that we are running into the same conceptual problem as with the previous solution: The file-creation and timestamp generation are not within the same critical section. Probably the filesystem first prepares all file meta-data (including the timestamp) and only then commits the changes to disk, allowing for a time window in which another process can run through the whole sequence of creating the file and checking all timestamps. Since this occurs within the same system call, the time between timestamp-creation and visibility of the new file is shorter, resulting in a lower probability of hitting this issue.

With your changes I see much reduced probability for this to occur. We initially ran into this after about 1 Month, while starting three concurrent jf-config processes at the same time once a day. Considering that in the tests the error now happened after more than 30min compared to the 10-20s it was before and with 4 times as many threads, we can assume at least a factor of 100x reduction in probability. On our CI server we have circumvented this issue now by assigning individual config files per worker, using the JFROG_CLI_HOME_DIR environment variable.

For correctness it would probably be best to use a mechanism explicitly designed for file locking (i.e. LockFileEx on windows and flock on linux). There is already a cross platform go implementation used for exactly the same usecase of protecting configfiles in the "go" commandline tool: [go.internal.lockedfile](https://pkg.go.dev/cmd/go/internal/lockedfile). Unfortunately this is an internal package. There is a project containing useful internal packages from the go standard library which [exports](https://github.com/rogpeppe/go-internal/tree/master/lockedfile) this though: https://github.com/rogpeppe/go-internal

common/commands/config.go
```diff
@@ -71,7 +80,34 @@ func (cc *ConfigCommand) SetDetails(details *config.ServerDetails) *ConfigComman
...
+	case AddOrEdit:
+		return cc.config()
+	case Delete:
+		return cc.delete()
+	case Use:
+		return cc.use()
+	case Clear:
+		return cc.clear()
+	default:
+		return fmt.Errorf("Not supported config command type: " + string(cc.cmdType))
```
eyalbe4, line 109: Not supported... --> Unsupported...

common/commands/config.go
```diff
@@ -211,7 +234,7 @@ func (cc *ConfigCommand) prepareConfigurationData() ([]*config.ServerDetails, er
 	return configurations, err
 }
 
-/// Returning the first non empty value:
+/// Returning the first non-empty value:
```
eyalbe4, line 237: ///  --> //

eyalbe4 (changes requested):

PR #354, opened 2022-03-15
Return error in config export if there's no default

asafgabai:
-----


common/commands/config.go
```diff
@@ -396,6 +396,9 @@ func Export(serverName string) error {
 	if err != nil {
 		return err
 	}
+	if serverDetails.ServerId == "" {
+		return errorutils.CheckErrorf("couldn't find default server")
```
eyalbe4, line 400: couldn't find default server
-->
Cannot export config, becuase it is empty. Run 'jf c add' and then export again

common/commands/config_test.go
```diff
@@ -131,6 +133,22 @@ func TestBasicAuthOnlyOption(t *testing.T) {
 	assert.NoError(t, DeleteConfig("test"))
 }
 
+func TestExportNoDefaultServer(t *testing.T) {
```
eyalbe4, line 136: TestExportNoDefaultServer --> TestExportEmptyConfig

eyalbe4 (changes requested):

PR #357, opened 2022-03-20
Dependencies in NuGet subprojects are mixed

yahavi:
-----
    
Resolves https://github.com/jfrog/jfrog-cli/issues/1281#issuecomment-1068892650
Test: https://github.com/jfrog/jfrog-cli/pull/1482

In NuGet projects, a possible scenario is that project A contains project B as a project reference (like sub-module).
Currently, all dependencies of B appear in A. Furthermore - the RequestedBy field of these dependencies is nil because there is no way to traverse them from A.

This PR suggests removing the unreachable dependencies.

### DIFF
--- a/artifactory/utils/dotnet/solution/solution.go
+++ b/artifactory/utils/dotnet/solution/solution.go
@@ -80,7 +80,11 @@ func (solution *solution) BuildInfo(moduleName string) (*buildinfo.BuildInfo, er
 
 		// Populate module dependencies
 		for _, dep := range dependencies {
-			module.Dependencies = append(module.Dependencies, *dep)
+			// If dependnecy has no RequestedBy field, it is not accessible in the current project -
+			// It is expected to be under a sub project.
+			if len(dep.RequestedBy) > 0 {
+				module.Dependencies = append(module.Dependencies, *dep)
+			}
 		}
 
 		modules = append(modules, module)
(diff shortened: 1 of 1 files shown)


artifactory/utils/dotnet/solution/solution.go
```diff
@@ -80,7 +80,11 @@ func (solution *solution) BuildInfo(moduleName string) (*buildinfo.BuildInfo, er
 
 		// Populate module dependencies
 		for _, dep := range dependencies {
-			module.Dependencies = append(module.Dependencies, *dep)
+			// If dependnecy has no RequestedBy field, it is not accessible in the current project -
```
eyalbe4, line 83: If dependnecy has no RequestedBy field, it is not accessible in the current project -
-->
If dependency has no RequestedBy field, it means that the depedency not accessible in the current project.

artifactory/utils/dotnet/solution/solution.go
```diff
@@ -80,7 +80,11 @@ func (solution *solution) BuildInfo(moduleName string) (*buildinfo.BuildInfo, er
...
+			// It is expected to be under a sub project.
```
eyalbe4, line 84: It is expected to be under a sub project.
-->
In that case, the dependency is assumed to be under a project which is referenced by this project. We therefore don't include the dependency in the build-info.

eyalbe4 (changes requested):

PR #364, opened 2022-03-28
Add invitation flow

gailazar300:
-----
* Add a new "invite" command which sends an invitation to the platform email.
* Change "setup" command- in case we want to setup the environment for an invited user, the setup command will get his credentials as a parameter and skip the registration step.


eyalbe4 (changes requested):
I haven't finished reviewing everything yet, but I think I'll soon go over everything.
I'm sharing what I have so far.

utils/config/config_test.go
```diff
@@ -57,9 +57,9 @@ func TestConvertConfigV0ToV5(t *testing.T) {
 	defer cleanUpTempEnv()
 	content, err := convertIfNeeded([]byte(configV0))
 	assert.NoError(t, err)
-	configV5 := new(ConfigV5)
-	assert.NoError(t, json.Unmarshal(content, &configV5))
-	assertionV5Helper(t, configV5, coreutils.GetCliConfigVersion(), false)
+	configV6 := new(ConfigV6)
+	assert.NoError(t, json.Unmarshal(content, &configV6))
+	assertionHelper(t, configV6, 0, false)
 	assertCertsMigrationAndBackupCreation(t)
```
eyalbe4, line 63: Shouldn't the name fo this function be changed from `TestConvertConfigV0ToV5` to `TestConvertConfigV0ToLatest`?
The same question applies to `TestConvertConfigV1ToV5` and `TestConvertConfigV4ToV5`.

utils/config/tokenrefresh.go
```diff
@@ -113,16 +143,36 @@ func tokenRefreshHandler(currentAccessToken string) (newAccessToken string, err
 		log.Debug("Token refreshed successfully.")
 	}
 
-	err = writeNewTokens(serverConfiguration, tokenRefreshServerId, newToken.AccessToken, newToken.RefreshToken)
+	err = writeNewArtifactoryTokens(serverConfiguration, tokenRefreshServerId, newToken.AccessToken, newToken.RefreshToken)
+	if err != nil {
+		log.Error("Failed writing new tokens to config after handling access token expiry: " + err.Error())
```
eyalbe4, line 148: I think that instead adding error message to the log, we should return here an error with this message.

utils/config/tokenrefresh.go
```diff
@@ -113,16 +143,36 @@ func tokenRefreshHandler(currentAccessToken string) (newAccessToken string, err
...
+	newToken, err := refreshExpiredAccessToken(serverConfiguration, currentAccessToken, refreshToken)
+
+	if err != nil {
+		log.Debug("Refresh access token failed: " + err.Error())
+		return "", err
+	}
+
+	err = writeNewArtifactoryTokens(serverConfiguration, tokenRefreshServerId, newToken.AccessToken, newToken.RefreshToken)
 	if err != nil {
 		log.Error("Failed writing new tokens to config after handling access token expiry: " + err.Error())
```
eyalbe4, line 168: I think that instead adding error message to the log, we should return here an error with this message.

utils/config/tokenrefresh.go
```diff
@@ -113,16 +143,36 @@ func tokenRefreshHandler(currentAccessToken string) (newAccessToken string, err
...
+func refreshAccessTokenAndWriteToConfig(serverConfiguration *ServerDetails, currentAccessToken string) (string, error) {
+	refreshToken := serverConfiguration.RefreshToken
+	// Remove previous tokens
+	serverConfiguration.AccessToken = ""
+	serverConfiguration.RefreshToken = ""
+	// Try refreshing tokens
+	newToken, err := refreshExpiredAccessToken(serverConfiguration, currentAccessToken, refreshToken)
+
+	if err != nil {
+		log.Debug("Refresh access token failed: " + err.Error())
```
eyalbe4, line 162: There's probably no need for thid debug log message. Instead, let's add this message to the error that is returned here.

eyalbe4 (changes requested):

utils/config/tokenrefresh.go
```diff
@@ -153,13 +203,14 @@ func createTokensForConfig(serverDetails *ServerDetails, expirySeconds int) (ser
...
 	if err != nil {
-		return services.CreateTokenResponseData{}, err
+		return auth.CreateTokenResponseData{}, err
 	}
 	return newToken, nil
 }
 
 func CreateInitialRefreshableTokensIfNeeded(serverDetails *ServerDetails) (err error) {
-	if !(serverDetails.TokenRefreshInterval > 0 && serverDetails.RefreshToken == "" && serverDetails.AccessToken == "") {
+	if !(serverDetails.ArtifactoryTokenRefreshInterval > 0 && serverDetails.RefreshToken == "" && serverDetails.AccessToken == "") ||
```
RobiNino, line 212: Seems to me like there's a mix up between the tokens, and I'm not sure why the second condition is needed. Care to explain?

RobiNino (changes requested):

RobiNino (changes requested):

utils/config/tokenrefresh.go
gailazar300: Yes, I got confused between the tokens - fixed it.

general/envsetup/envsetup_test.go
```diff
@@ -0,0 +1,16 @@
+package envsetup
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestSetupInvitedUser(t *testing.T) {
```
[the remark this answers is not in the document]
gailazar300: I can test it in cli's tests - cause if I will run the config command on this invalid base64 value it will fail (trying to ping artifactory in the end and fails all the config flow)
also can't generate from test's short lived accessToken a refreshToken. it can be tested only in integration tests.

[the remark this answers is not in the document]
eyalbe4: OK, so you're going to leave this test as is, and add an additional test on `jfrog-cli`?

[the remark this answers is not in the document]
gailazar300: yes

PR #367, opened 2022-03-29
Make all config commands - atomic

sverdlov93:
-----
Make all config commands that write to config file - atomic.
should fix https://github.com/jfrog/jfrog-cli-core/issues/350


artifactory/commands/generic/upload.go
```diff
@@ -174,6 +174,7 @@ func (uc *UploadCommand) upload() (err error) {
 			log.Error(err)
 		}
 	}
+	log.Info("Uploaded successfully!")
```
eyalbe4, line 177: I suggest removing this new log message, as it'll add one additional info message for each file upload. I'm not sure this is needed.

eyalbe4 (changes requested):

PR #374, opened 2022-04-14
Add optional 'resources' directory to plugins

gailazar300:
-----
* Adding an option to create and publish plugins to Artifactory with a directory named "resources", which contains all resources that the developer wants to add to the plugin.
* When publishing a plugin make sure that you have a "resources" directory in root.
* Resources are being uploaded to Artifactory in a zip file.
* There are changes in the plugins directory hierarchy at the local '.jfrog' directory.


utils/plugins/utils.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		return err
+	}
+	if plugins.Version != coreutils.GetPluginsYamlVersion() {
+		return errorutils.CheckError(errors.New(fmt.Sprintf("Expected plugins version in 'plugins.yaml is %d  but the actual value is %d", coreutils.GetPluginsYamlVersion(), plugins.Version)))
+	}
+	return nil
+}
+
+func readPluginsYaml() (*PluginsV1, error) {
```
eyalbe4, line 67: readPluginsYaml --> readPluginsConfig

utils/plugins/utils.go
```diff
@@ -0,0 +1,202 @@
...
+	if !exists {
+		return nil
+	}
+
+	plugins, err := readPluginsYaml()
+	if err != nil {
+		return err
+	}
+	if plugins.Version != coreutils.GetPluginsYamlVersion() {
+		return errorutils.CheckError(errors.New(fmt.Sprintf("Expected plugins version in 'plugins.yaml is %d  but the actual value is %d", coreutils.GetPluginsYamlVersion(), plugins.Version)))
```
eyalbe4, line 62: Two white-spaces before "but", when only one is needed.

utils/plugins/utils.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	err = json.Unmarshal(content, &plugins)
+	if err != nil {
+		return nil, errorutils.CheckError(err)
+	}
+	return plugins, err
+}
+
+func getPluginsYamlFile() (content []byte, err error) {
```
eyalbe4, line 86: getPluginsYamlFile --> getPluginsConfigFileContent

utils/plugins/utils.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func readPluginsYaml() (*PluginsV1, error) {
+	plugins := new(PluginsV1)
+	content, err := getPluginsYamlFile()
+	if err != nil {
+		return nil, err
+	}
+	if len(content) == 0 {
+		// No plugins.yaml file was found, that means that we are in v0.
```
eyalbe4, line 74: ...found, that... --> ...found. This...

utils/plugins/utils.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		return
+	}
+	if exists {
+		content, err = fileutils.ReadFile(pluginsFilePath)
+	}
+	return
+}
+
+// V0: in plugins directory there was no 'plugins.yaml' file. all executable files were in the same directory.
```
eyalbe4, line 101: ...in plugins directory... --> ...In the plugins directory...
...all executable... --> ...This means that all executable...
were --> are

utils/plugins/utils.go
```diff
@@ -0,0 +1,202 @@
...
+// V1: create 'plugins.yaml' file inside 'plugins' directory. change the file's hierarchy inside 'plugins' directory.
```
eyalbe4, line 102: // V1: create 'plugins.yaml' file inside 'plugins' directory. change the file's hierarchy inside 'plugins' directory.
-->
// V1: We should create a 'plugins.yml' file inside the 'plugins' directoryand also modify the files's hierarchy inside 'plugins' directory.

utils/plugins/utils.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+// Change the file's hierarchy inside 'plugins' directory to:
+//	plugins (dir)
+//		plugin-name (dir)
+//			bin (dir)
+//				plugin-executable (file)
+//			resources:(optional dir)
+//				... (directories/files)
+func migrateFileSystemLayoutV0ToV1() error {
```
eyalbe4, line 118: migrateFileSystemLayoutV0ToV1 --> convertFileSystemLayoutV0ToV1
(since we're already using the tern "convert" in the code, we better be consistent and not use "migrate".

utils/plugins/utils.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	pluginsDir, err := coreutils.GetJfrogPluginsDir()
+	for _, p := range plugins {
+		// Skip 'plugins.yaml'
+		if p.Name() == coreutils.JfrogPluginsFile {
+			continue
+		}
+		if p.IsDir() {
+			log.Error("unexpected directory in plugins directory")
+			break
```
eyalbe4, line 134: Don't we prefer to continue here, instead of break?

utils/plugins/utils.go
```diff
@@ -0,0 +1,202 @@
...
+		return err
+	}
+	pluginsDir, err := coreutils.GetJfrogPluginsDir()
+	for _, p := range plugins {
+		// Skip 'plugins.yaml'
+		if p.Name() == coreutils.JfrogPluginsFile {
+			continue
+		}
+		if p.IsDir() {
+			log.Error("unexpected directory in plugins directory")
```
eyalbe4, line 133: Let's also add to the error message the full path of the unexpected directory.

utils/plugins/utils.go
```diff
@@ -0,0 +1,202 @@
...
+			break
+		}
+		// Verify that the file is an executable file
+		if !IsExec(p.Mode()) {
+			log.Error("unexpected file in plugins directory: " + p.Name())
```
eyalbe4, line 138: Let's be more specific and clear than this error message. The user will not understand why is the file "unexpected".

utils/plugins/utils.go
```diff
@@ -0,0 +1,202 @@
...
+			continue
+		}
+		// Move plugins exec files inside a directory with the plugin's name.
```
eyalbe4, line 141: ... directory with the plugin's name. --> ...directory, which has the plugin's name.

utils/plugins/utils.go
```diff
@@ -0,0 +1,202 @@
...
+		// Create a directory with the plugin's name + "_dir" extension, move the file inside and change directory's name back to plugin's name only.
+		pluginsName := getPluginsNameFromExec(p.Name())
+		err = os.MkdirAll(filepath.Join(pluginsDir, pluginsName+"_dir", coreutils.PluginsExecDirName), 0777)
+		if err != nil {
+			return err
```
eyalbe4, line 150: Missing CheckError here.

utils/plugins/utils.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+		err = fileutils.MoveFile(filepath.Join(pluginsDir, p.Name()), filepath.Join(pluginsDir, pluginsName+"_dir", coreutils.PluginsExecDirName, p.Name()))
+		if err != nil {
+			return err
+		}
+		err = os.Rename(filepath.Join(pluginsDir, pluginsName+"_dir"), filepath.Join(pluginsDir, pluginsName))
```
eyalbe4, line 160: We're noy allowed to use the ```os.Rename``` function, because ti does not work inside docker containers. See more info in the documentation of the ```fileutils.MoveFile``` function.

utils/plugins/utils.go
```diff
@@ -0,0 +1,202 @@
...
+		if err != nil {
+			return err
```
eyalbe4, line 162: Missing CheckError here,

utils/plugins/utils.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	return nil
+}
+
+// IsExec Check if file's type is executable
+func IsExec(mode os.FileMode) bool {
+	return mode&0111 != 0
+}
+
+func getPluginsNameFromExec(execName string) string {
```
eyalbe4, line 173: getPluginsNameFromExec --> removeFileExtension
execName --> fileName

utils/plugins/utils.go
```diff
@@ -0,0 +1,202 @@
...
+	return strings.Split(execName, ".")[0]
+}
+
+func CreatePluginsYamlFile() (*PluginsV1, error) {
```
eyalbe4, line 177: CreatePluginsYamlFile --> createPluginsConfigFile

utils/plugins/utils.go
```diff
@@ -0,0 +1,202 @@
...
+	pluginsFilePath, err := getPluginsFilePath()
+	if err != nil {
+		return nil, err
+	}
+	plugins := new(PluginsV1)
+	plugins.Version = 1
```
eyalbe4, line 183: There's already a function that returns the value "1". Let's use it instead of having this value in the code once more. This will make it easier when moving to version 2.

utils/plugins/utils.go
```diff
@@ -0,0 +1,202 @@
...
+	content, err := json.Marshal(plugins)
+	if err != nil {
+		return nil, errorutils.CheckError(err)
+	}
+	err = ioutil.WriteFile(pluginsFilePath, content, 0777)
```
eyalbe4, line 188: 0777 is too strong. The file should have only read and wrte permissions, but not execute permissions.

utils/plugins/utils_test.go
```diff
@@ -0,0 +1,103 @@
...
+	exists, err := fileutils.IsFileExists(pluginsExecName, false)
+	if exists {
+		err = os.Chmod(pluginsExecName, 0777)
+		assert.NoError(t, err)
+	}
+	return testHomeDir
+}
+
+// Set JFROG_CLI_HOME_DIR environment variable to be a new temp directory
+func CreateTempEnvForPluginsTests(t *testing.T) (cleanUp func()) {
```
eyalbe4, line 90: Looks like this function can be made private.

Also, it looks like we're usinbg a real 18mb sized plugin for the testing. Can we use anither executable, which is not a real plugin,so that we don't store such big files in git?

eyalbe4 (changes requested):

PR #380, opened 2022-04-19
Add a build-info UI link to the output of the Build Publish command

RobiNino:
-----


artifactory/formats/buildpublishapi.go
```diff
@@ -0,0 +1,19 @@
+package formats
+
+import (
+	"bytes"
+	"encoding/json"
+)
+
+type BuildPublishOutput struct {
```
eyalbe4, line 8: I suggets merging the content of thsi file into the ```buildinfo/publish.go``` file, becuase it seems to be used by the build info publish command only.

eyalbe4 (changes requested):

artifactory/formats/buildpublishapi.go
RobiNino: @eyalbe4 I created this file to mark that any change to this struct will break the API. Added a comment that clarifies it.
We discussed it before and agreed that this is the preferred way to protect APIs used externally (used by Azure in this case). Let me know what you think

eyalbe4: OK. NP.

eyalbe4 (approved):

Issue #382, opened 2022-04-26
.NET Project dependencies not found when project is not located under solution directory


prochnowc:
**Describe the bug**
If the .NET project file is not located in a directory below the solution file no dependencies are found.

**To Reproduce**
Create a project with the following directory layout:
- tmp\solution.sln
- project1\project1.csproj

Run the JFrog CLI to record the dependencies when packages are restored. "jfrog dotnet restore tmp\solution.sln --build-name test --build-number 1"
project1 is reported with "Project dependencies was not found for project"

**Expected behavior**
I expect that dependencies are found.

**Versions**
- JFrog CLI version (if applicable): 2.15.1

eyalbe4:
@prochnowc ,
Thanks for reporting this issue.
Will you be able to share a smaple project on GitHub that reproduces this issue?

prochnowc:
[jfrog-cli-issue.zip](https://github.com/jfrog/jfrog-cli-core/files/8560423/jfrog-cli-issue.zip)

To reproduce change to the "jfrog-cli-issue" directory and run the following commands:

```
$env:JFROG_CLI_LOG_LEVEL="DEBUG"
jfrog dotnet-config
jfrog dotnet restore .\solutions\test.sln --build-name test --build-number 1
```

You will see the message `[Debug] Project dependencies was not found for project: project1` and dependencies are not recorded in the build info.

BTW.. i'm running this on Windows, dont know if linux works.

prochnowc:
@eyalbe4 I want to mention that this issue is blocking us from obtaining proper build info's. We have a pro-x license for Artifactory. Thank you.

eyalbe4:
@prochnowc,
Just updating that this task is included in our work plan, We'll share here an update once this is implemented.

prochnowc:
@eyalbe4 Thanks for fixing this issue. May I ask when it gets released?

gailazar300:
@prochnowc Jfrog CLI v2.18.0 is out with this issue fix.
We would appreciate it if you could give it a try, and let us know if it works as aspected.

prochnowc:
@gailazar300 Thanks for the fix. It works now.

PR #390, opened 2022-05-18
Fix: .NET dependencies not found when project is not located under solution directory

gailazar300:
-----
Fix issue: https://github.com/jfrog/jfrog-cli-core/issues/382
When:
1. Running 'jf dotnet resolve <solution path>' command.
2. Projects are not located under the solution directory. 
We get the "Project dependencies was not found for project" message.


artifactory/utils/dotnet/solution/solution.go
```diff
@@ -141,19 +147,34 @@ func (solution *solution) GetProjects() []project.Project {
...
 	if slnProjects != nil {
-		return solution.loadProjectsFromSolutionFile(slnProjects)
+		return solution.parseProjectsFromSolutionFile(slnProjects)
 	}
+	return nil, nil
+}
 
-	return solution.loadSingleProjectFromDir()
+func (solution *solution) loadProjects(slnProjects []project.Project) error {
+	// No '.sln' file was provided as a parameter/founded - load project from the given directory.
```
RobiNino, line 162: founded -> found

artifactory/utils/dotnet/solution/solution.go
```diff
@@ -181,18 +202,19 @@ func (solution *solution) loadSingleProjectFromDir() error {
 
 	if len(projFiles) == 1 {
```
RobiNino, line 203: We might want to log something here, because the cli will return without collecting build info and without throwing an error.

RobiNino (changes requested):

PR #421, opened 2022-06-21
Get repository names with filters

asafgabai:
-----
Get all repository names, filtered by their names and types.


artifactory/utils/repositoryutils.go
```diff
@@ -98,3 +69,60 @@ func IsRemoteRepo(repoName string, serviceManager artifactory.ArtifactoryService
...
+func GetFilteredRepositories(servicesManager artifactory.ArtifactoryServicesManager, includePatterns, excludePatterns []string, repoType *RepoType) ([]string, error) {
+	filterParams := services.RepositoriesFilterParams{RepoType: repoType.String()}
+	repos, err := servicesManager.GetAllRepositoriesFiltered(filterParams)
+	if err != nil {
+		return nil, err
+	}
+	return filterRepositoryNames(repos, includePatterns, excludePatterns)
+}
+
+func filterRepositoryNames(repos *[]services.RepositoryDetails, includePatterns, excludePatterns []string) ([]string, error) {
```
eyalbe4, line 89: I think that the logic in thsi function can optimized as follows.
If no include patterns were sent, we can avoid running `filepath.Match` on the list of repositories. The same goes for an empty exclude patterns list. This is important because this API is expected to be used with hundreds or even thousands of repositories, and the `filepath.Match` isn't trivial in terms of cost.

eyalbe4 (changes requested):

PR #426, opened 2022-06-24
Refine usage of defer in config

arcanericky:
-----
Refine atomic config command changes in `common/commands/ConfigCommand.Run()`, int https://github.com/jfrog/jfrog-cli-core/pull/367, and also mentioned in https://github.com/jfrog/jfrog-cli-core/pull/352.

Change Details

- The unlock and debug logging are more clean (IMO) when performed in an anonymous function rather than stacking `defer` statements when possible. Note that I've accounted for `defer` executing in LIFO order.
- The `defer` of the anonymous functon containing the `lockFile.Unlock()` gets executed _after_ the return but _before_ the function returns to the caller (https://go.dev/ref/spec#Defer_statements). This means the setting and check for `e` not being `nil` and setting `err` have no effect on what is returned as might be the intention. See below for more. This change refines the code while leaving the current behavior in place because I know a lot of work went into shoring up the file lock race conditions. If returning an error from `lockFile.Unlock()` is the intention, more invasive changes must be made, and this isn't the only usage of this pattern (again, see below).
- My `gofmt` that's part of `1.17.11` _really_ wants to reorganize the imports so I left those changes in since the README explicity states to use `gofmt`.
- Fixed an extra `/` in the comment as mentioned in https://github.com/jfrog/jfrog-cli-core/pull/352#discussion_r872979948.

For bullet point number 2, I see this pattern scattered throughout the code base:

```go
defer func() {
  e := funcThatCanReturnError()
  if err == nil {
    err = e
  }
}()
```

Where it is expected that when the surrounding function returns `err`, could contain the `err` value set in the `defer`. This is not correct. You can observe the true behavior by running the following code which I've also placed in the Go Playground at https://go.dev/play/p/gOQB-jPRu4B:

```go
func deferTest() error {
  err := errors.New("top")
  defer func() { err = errors.New("defer") }()
  return err
}

func deferPatternTest() error {
  var err error = nil
  defer func() {
    e := func() error { return errors.New("defer") }()
    if err == nil {
      err = e
    }
  }()
  return err
}

func main() {
  fmt.Println(deferTest())
  fmt.Println(deferPatternTest())
}
```

which outputs

```shell
$ go run deferdemo.go 
top
<nil>
```

### DIFF
--- a/common/commands/config.go
+++ b/common/commands/config.go
@@ -91,12 +91,11 @@ func (cc *ConfigCommand) Run() (err error) {
 
 	lockDirPath, err := coreutils.GetJfrogConfigLockDir()
 	if err != nil {
-		return
+		return err
 	}
-
 	lockFile, err := lock.CreateLock(lockDirPath)
 	if err != nil {
-		return
+		return err
 	}
 	defer func() {
 		e := lockFile.Unlock()
@@ -107,18 +106,16 @@ func (cc *ConfigCommand) Run() (err error) {
 
 	switch cc.cmdType {
 	case AddOrEdit:
-		err = cc.config()
+		return cc.config()
 	case Delete:
-		err = cc.delete()
+		return cc.delete()
 	case Use:
-		err = cc.use()
+		return cc.use()
 	case Clear:
-		err = cc.clear()
+		return cc.clear()
 	default:
-		err = fmt.Errorf("Not supported config command type: " + string(cc.cmdType))
+		return fmt.Errorf("Not supported config command type: " + string(cc.cmdType))
 	}
-
-	return
 }
 
 func (cc *ConfigCommand) ServerDetails() (*config.ServerDetails, error) {
--- a/common/commands/config.go
+++ b/common/commands/config.go
@@ -2,14 +2,15 @@ package commands
 
 import (
 	"fmt"
-	"github.com/jfrog/jfrog-cli-core/v2/utils/coreutils"
-	"github.com/jfrog/jfrog-cli-core/v2/utils/ioutils"
-	"github.com/jfrog/jfrog-cli-core/v2/utils/lock"
 	"io/ioutil"
 	"reflect"
 	"strconv"
 	"strings"
 	"sync"
+
+	"github.com/jfrog/jfrog-cli-core/v2/utils/coreutils"
+	"github.com/jfrog/jfrog-cli-core/v2/utils/ioutils"
+	"github.com/jfrog/jfrog-cli-core/v2/utils/lock"
 
 	"github.com/jfrog/jfrog-client-go/auth"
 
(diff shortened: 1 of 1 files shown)


arcanericky:

Or-Geva:
@arcanericky, thanks for your great investigation. It's much appreciated!.
Regarding your bullet point number 2, I believe the right fix will require naming the return value so that the _error catch_ would work.
I've implemented it on the playground demo [here](https://go.dev/play/p/k2ladwILtyx).

arcanericky:
Yes, your solution works great and is simpler than I thought it would be. I'm sure it was overlooked here because in at least a few locations I see where your solution is coded and avoids the problem, though I have found a few more "opportunities".

In the case of `ConfigCommand.Run()`, more changes were required because the results of function calls are immediately returned. [This playground demo](https://go.dev/play/p/BHhjeUFueOv) matches the code in `ConfigCommand.Run()` a bit more. My fix is to assign the result of the function calls to `err` and `return` at the end. For readability, I like removing `err` from the `return` statements. I've pushed my changes with these additions.

arcanericky:
It's important when you review this to keep in mind the original intention may be to have the return values from the function calls in the `switch` _override_ any error that may occur with the `lockFile.Unlock()`. Being an ourside observer, it's tough for me to tell with no comments here.

If this is the case, you should dismiss this PR or have me withdraw it.

arcanericky:
I've reverted to allowing the function calls in the `switch` to override any setting of `err` in the `defer`. I refuse to be responsible for breaking current behavior (except the origin `defer` problem), intended or not.

common/commands/config.go
```diff
@@ -79,25 +80,30 @@ func (cc *ConfigCommand) SetDetails(details *config.ServerDetails) *ConfigComman
...
 
-func (cc *ConfigCommand) Run() error {
+func (cc *ConfigCommand) Run() (err error) {
 	log.Debug("Locking config file to run config " + cc.cmdType + " command.")
-	defer log.Debug("Config " + cc.cmdType + " command completed successfully. config file is released.")
 	mutex.Lock()
-	defer mutex.Unlock()
+
+	defer func() {
+		mutex.Unlock()
```
Or-Geva, line 88: Let's not ignore this error

sverdlov93:
Hi, @arcanericky.
Agree with the first bullet, adjacent defers should be indeed collected together inside an anonymous function.
Regarding the second bullet, the reason behind this order is to avoid deadlock issues.
If somehow the lock would be created and then some error would happen, we must add the unlock function to our defer calls before throwing the error.
Combining the previous order with the `(error err)` convention will achieve both goals here. 

Here is the changed code:
```
func (cc *ConfigCommand) Run() (err error) {
	log.Debug("Locking config file to run config " + cc.cmdType + " command.")
	mutex.Lock()
	defer func() {
		mutex.Unlock()
		log.Debug("Config " + cc.cmdType + " command completed successfully. config file is released.")
	}()

	lockDirPath, err := coreutils.GetJfrogConfigLockDir()
	if err != nil {
		return
	}
	lockFile, err := lock.CreateLock(lockDirPath)
	defer func() {
		e := lockFile.Unlock()
		if err == nil {
			err = e
		}
	}()
	if err != nil {
		return
	}

	switch cc.cmdType {
	case AddOrEdit:
		err = cc.config()
	case Delete:
		err = cc.delete()
	case Use:
		err = cc.use()
	case Clear:
		err = cc.clear()
	default:
		err = fmt.Errorf("Not supported config command type: " + string(cc.cmdType))
	}
	return
}
```

arcanericky:
Your PR looks good @sverdlov93. One of you can close this PR or I will shut it down at your request.

sverdlov93:
@arcanericky Thanks again for your contribution and your feedback! 
Once my PR will be reviewed I will merge it and close that one.

PR #437, opened 2022-06-29
Improve getting transfer errors files

RobiNino:
-----


artifactory/commands/transferfiles/utils.go
```diff
@@ -264,13 +267,22 @@ func getErrorsFiles(repoKey string, isRetry bool) (filesPaths []string, err erro
 		return []string{}, err
 	}
 
-	filesNames, err := biUtils.ListFiles(dirPath, false)
+	errorsFilesRegex := fmt.Sprintf(errorsFilesRegexFormat, repoKey)
+
+	regExp, err := regexp.Compile(errorsFilesRegex)
```
eyalbe4, line 272: ```regexp.Compile``` is an expensive operation. It is therefore better to it only once per CLI execution.

artifactory/commands/transferfiles/utils.go
```diff
@@ -264,13 +267,22 @@ func getErrorsFiles(repoKey string, isRetry bool) (filesPaths []string, err erro
...
 	if err != nil {
 		return nil, err
 	}
 
-	for _, file := range filesNames {
-		if strings.HasPrefix(filepath.Base(file), repoKey) {
+	for _, file := range files {
+		matches := regExp.FindStringSubmatch(filepath.Base(file))
+		// Expecting a match and 4 groups.
+		if len(matches) == 5 {
```
eyalbe4, line 285: We're checking that the len is 5, but the above comment says `4 groups`. Is this a mistake?

eyalbe4 (changes requested):

artifactory/commands/transferfiles/utils.go
RobiNino: No, this function returns the whole match and the value of each group. I'll try rephrasing the comment above.

artifactory/commands/transferfiles/utils.go
RobiNino: This regexp contains the repo key, so it cannot be compiled once. As discussed, leaving as it is

PR #443, opened 2022-07-01
Transfer restructure, handle deploy order

RobiNino:
-----


artifactory/commands/transferfiles/delayedartifactshandler.go
```diff
@@ -0,0 +1,209 @@
...
+	"path"
+	"path/filepath"
+)
+
+const (
+	// Max delayedArtifacts that will be written in a file
+	maxDelayedArtifactsInFile = 50000
+)
+
+// TransferDelayedArtifactsMng managing multi threads writing delayed deployment artifacts in order to keep the order of deployment.
```
eyalbe4, line 19: // TransferDelayedArtifactsMng managing multi threads writing delayed deployment artifacts in order to keep the order of deployment.
-->
// TransferDelayedArtifactsMng takes care of the multi threaded writing of artifacts to be transferred. while maintaining the correct order of the deployment. This is needed because, for example, for maven repositories, pom file should be deployed last.

artifactory/commands/transferfiles/delayedartifactshandler.go
```diff
@@ -0,0 +1,209 @@
...
+
+type delayedArtifactWriter struct {
+	writer               *content.ContentWriter
+	delayedArtifactCount int
+	// In case we have multiple delayedArtifacts files - we index them
+	fileIndex int
+}
+
+func (mng *TransferDelayedArtifactsMng) initDelayedArtifactWriter() (err error) {
+	// Init the content writer which responsible for writing delayed artifacts - That means delayed artifacts that we should upload later, in a certain order.
```
eyalbe4, line 37: ...writer which responsible... --> ...writer, which is responsible...

... That means delayed artifacts that we should upload later, in a certain order.
-->
... This means that delayed artifacts will be deployed later, to maintain the right deployment order.

artifactory/commands/transferfiles/delayedartifactshandler.go
```diff
@@ -0,0 +1,209 @@
...
+	writer, err := mng.newContentWriter()
```
eyalbe4, line 38: The fact that we create the writer here means that we will close it in a different function. This is not advisable, because it makes it hard to track where exactly it will be closed.
The right approach should be -
1. Create the writer
2. Close it inside a defer 

This may mean that would need to create and close it somewhere else. Maybe the "start" function could be a good place.

artifactory/commands/transferfiles/delayedartifactshandler.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+func (mng *TransferDelayedArtifactsMng) start() error {
+	for file := range mng.delayedArtifactsChannel {
+		err := mng.writeDelayedArtifactContent(file)
+		if err != nil {
+			return err
+		}
+	}
+	return mng.closeWriter()
```
eyalbe4, line 60: As mentioned in a previos comment, this should happen ina defer, and also inside the dame function which created the writer.

artifactory/commands/transferfiles/delayedartifactshandler.go
```diff
@@ -0,0 +1,209 @@
...
+	return []shouldDelayUpload{}
+}
+
+type delayUploadHelper struct {
+	shouldDelayFunctions    []shouldDelayUpload
+	delayedArtifactsChannel chan FileRepresentation
+}
+
+// Decide whether to delay the deployment of a file by running over the shouldDelayUpload array.
+// This is an array to allow removing a single comparison when its delay is no longer needed.
```
eyalbe4, line 200: I think part of the description should be rephrased, becuase it isn't clear what "its" referees to - 
...when its delay is no longer needed.

artifactory/commands/transferfiles/manager.go
```diff
@@ -0,0 +1,153 @@
...
+
+func newTransferManager(base phaseBase, delayUploadComparisonFunctions []shouldDelayUpload) transferManager {
+	return transferManager{phaseBase: base, delayUploadComparisonFunctions: delayUploadComparisonFunctions}
+}
+
+type transferActionType func(optionalPcDetails producerConsumerDetails, uploadTokensChan chan string, delayHelper delayUploadHelper) error
+
+// This function handles a transfer process as part of a phase.
+// As part of the process, the transferAction gets executed. It may utilize a producer consumer or not.
+// The transferAction will collect artifacts to be uploaded into chunks, and sends them to the source Artifactory instance to handle.
```
eyalbe4, line 23: ...transferAction will collect artifacts... --> ...transferAction collects artifacts...

eyalbe4 (changes requested):

PR #447, opened 2022-07-03
Transfer - add pagination to full-transfer phase

RobiNino:
-----

### DIFF
--- a/artifactory/commands/transferfiles/filesdiff.go
+++ b/artifactory/commands/transferfiles/filesdiff.go
@@ -217,7 +217,7 @@ func (f *filesDiffPhase) getTimeFrameFilesDiff(repoKey, fromTimestamp, toTimesta
 }
 
 func generateDiffAqlQuery(repoKey, fromTimestamp, toTimestamp string, paginationOffset int) string {
-	query := fmt.Sprintf(`items.find({"type":"file","modified":{"$gte":"%s"},"modified":{"$lt":"%s"},"$or":[{"$and":[{"repo":"%s","path":{"$match":"*"},"name":{"$match":"*"}}]}]})`, fromTimestamp, toTimestamp, repoKey)
+	query := fmt.Sprintf(`items.find({"$and":[{"modified":{"$gte":"%s"}},{"modified":{"$lt":"%s"}},{"repo":"%s","path":{"$match":"*"},"name":{"$match":"*"}}]})`, fromTimestamp, toTimestamp, repoKey)
 	query += `.include("repo","path","name","modified")`
 	query += fmt.Sprintf(`.sort({"$asc":["modified"]}).offset(%d).limit(%d)`, paginationOffset*aqlPaginationLimit, aqlPaginationLimit)
 	return query
--- a/artifactory/commands/transferfiles/filesdiff.go
+++ b/artifactory/commands/transferfiles/filesdiff.go
@@ -164,23 +164,34 @@ func (f *filesDiffPhase) handleTimeFrameFilesDiff(params timeFrameParams, logMsg
 	toTimestamp := params.fromTime.Add(searchTimeFramesMinutes * time.Minute).Format(time.RFC3339)
 	log.Debug(logMsgPrefix + "Searching time frame: '" + fromTimestamp + "' to '" + toTimestamp + "'")
 
-	result, err := f.getTimeFrameFilesDiff(params.repoKey, fromTimestamp, toTimestamp)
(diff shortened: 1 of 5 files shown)


artifactory/commands/transferfiles/filesdiff.go
```diff
@@ -200,15 +211,16 @@ func convertResultsToFileRepresentation(results []servicesUtils.ResultItem) (fil
...
+	query := generateDiffAqlQuery(repoKey, fromTimestamp, toTimestamp, paginationOffset)
 	return runAql(f.srcRtDetails, query)
 }
 
-func generateDiffAqlQuery(repoKey, fromTimestamp, toTimestamp string) string {
-	items := fmt.Sprintf(`items.find({"type":"file","modified":{"$gte":"%s"},"modified":{"$lt":"%s"},"$or":[{"$and":[{"repo":"%s","path":{"$match":"*"},"name":{"$match":"*"}}]}]})`, fromTimestamp, toTimestamp, repoKey)
-	items += `.include("repo","path","name")`
-	return items
+func generateDiffAqlQuery(repoKey, fromTimestamp, toTimestamp string, paginationOffset int) string {
+	query := fmt.Sprintf(`items.find({"type":"file","modified":{"$gte":"%s"},"modified":{"$lt":"%s"},"$or":[{"$and":[{"repo":"%s","path":{"$match":"*"},"name":{"$match":"*"}}]}]})`, fromTimestamp, toTimestamp, repoKey)
```
yahavi, line 220:
```suggestion
	query := fmt.Sprintf(`items.find({"$and":[{"modified":{"$gte":"%s"}},{"modified":{"$lt":"%s"}},{"repo":"%s","path":{"$match":"*"},"name":{"$match":"*"}}]})`, fromTimestamp, toTimestamp, repoKey)
```
After reformatting the AQL, I got this JSON:
```json
{
  "type": "file",
  "modified": { "$gte": "%s" },
  "modified": { "$lt": "%s" },
  "$or": [
    {
      "$and": [
        { "repo": "%s", "path": { "$match": "*" }, "name": { "$match": "*" } }
      ]
    }
  ]
}
```
I have 2 comments:
1. Modified is duplicate and should be inside "and".
2. Redundant "or" clause.
3. The default type is already "file".

I suggest simplifying into this:
```json
{
  "$and": [
    { "modified": { "$gte": "%s" } },
    { "modified": { "$lt": "%s" } },
    { "repo": "%s", "path": { "$match": "*" }, "name": { "$match": "*" } }
  ]
}
```

artifactory/commands/transferfiles/fulltransfer.go
```diff
@@ -137,68 +137,77 @@ func (m *fullTransferPhase) transferFolder(params folderParams, logMsgPrefix str
...
-	}
 
-	// Empty folder. Add it as candidate.
-	if len(result.Results) == 0 {
-		curUploadChunk.appendUploadCandidate(FileRepresentation{
-			Repo: params.repoKey,
-			Path: path.Dir(params.relativePath),
-			Name: path.Base(params.relativePath),
-		})
+		if len(result.Results) < aqlPaginationLimit {
```
yahavi, line 202: Shouldn't be `<=`?

artifactory/commands/transferfiles/transfer.go
```diff
@@ -12,11 +12,12 @@ import (
 	"github.com/jfrog/jfrog-client-go/utils/errorutils"
 	"github.com/jfrog/jfrog-client-go/utils/log"
 	"os"
+	"strconv"
 )
 
 const (
 	tasksMaxCapacity = 10000
-	uploadChunkSize  = 100
+	uploadChunkSize  = 10
```
yahavi, line 20: After https://github.com/jfrog/data-transfer/pull/12, we can keep it large. Your decision.

yahavi (changes requested):

artifactory/commands/transferfiles/fulltransfer.go
RobiNino: No, if `len(result.Results) == aqlPaginationLimit` we don't know if there are more results but they were not returned because of the limit. This will make this loop run once only.
When `len(result.Results) < aqlPaginationLimit` we know we can stop because the results weren't truncated by the limit.

artifactory/commands/transferfiles/transfer.go
RobiNino: Agreed.

PR #450, opened 2022-07-04
Stop and exit JFrog CLI if one of the transfer's writers fails to write 

gailazar300:
-----
In case of getting errors while:
1. Write the transfer's errors statuses into the errors files (located in the JFrog CLI home directory).
2. Write the transfer's delayed artifacts into files (files that should be deployed to artifactory after other regular files)
In those cases- stop the transfer process.

### DIFF
--- a/artifactory/commands/transferfiles/errorshandler.go
+++ b/artifactory/commands/transferfiles/errorshandler.go
@@ -89,11 +89,7 @@ func initTransferErrorsDir() error {
 	if err != nil {
 		return err
 	}
-	err = makeDirIfDoesNotExists(skipped)
-	if err != nil {
-		return err
-	}
-	return nil
+	return makeDirIfDoesNotExists(skipped)
 }
 
 func makeDirIfDoesNotExists(path string) error {
@@ -110,7 +106,7 @@ func makeDirIfDoesNotExists(path string) error {
 func (mng *TransferErrorsMng) start() (err error) {
 	// Init content writers manager
 	writerMng := errorWriterMng{}
-	// Init the content writer which responsible for writing 'retryable errors' into files.
+	// Init the content writer which is responsible for writing 'retryable errors' into files.
 	// In the next run we would like to retry and upload those files again.
 	retryablePath, err := coreutils.GetJfrogTransferRetryableDir()
 	if err != nil {
@@ -127,7 +123,7 @@ func (mng *TransferErrorsMng) start() (err error) {
 		}
 	}()
 	writerMng.retryable = errorWriter{writer: writerRetry, fileIndex: 0, filePath: retryFilePath}
-	// Init the content writer which responsible for writing 'skipped errors' into files.
+	// Init the content writer which is responsible for writing 'skipped errors' into files.
 	// In the next run we won't retry and upload those files.
 	skippedPath, err := coreutils.GetJfrogTransferSkippedDir()
 	if err != nil {
@@ -323,7 +319,7 @@ func getErrorsFiles(repoKey string, isRetry bool) (filesPaths []string, err erro
 	return
 }
 
-// ErrorsChannelMng handles the uploading errors and add them to a common channel.
+// ErrorsChannelMng handles the uploading errors and adds them to a common channel.
 // Stops adding elements to the channel if an error occurs while handling the files.
 type ErrorsChannelMng struct {
 	channel chan FileUploadStatusResponse


artifactory/commands/transferfiles/manager.go
```diff
@@ -76,15 +76,15 @@ func (ftm *transferManager) doTransfer(isProducerConsumer bool, transferAction t
...
 	}()
 
 	// Transfer action to execute.
 	runWaitGroup.Add(1)
 	var actionErr error
 	go func() {
 		defer runWaitGroup.Done()
-		actionErr = transferAction(pcDetails, uploadTokensChan, delayUploadHelper{shouldDelayFunctions: ftm.delayUploadComparisonFunctions, delayedArtifactsChannel: delayedArtifactsChannel}, errorChannel)
+		actionErr = transferAction(pcDetails, uploadTokensChan, delayUploadHelper{shouldDelayFunctions: ftm.delayUploadComparisonFunctions, delayedArtifactsChannelMng: &delayedArtifactsChannelMng}, &errorsChannelMng)
 		if !isProducerConsumer {
```
eyalbe4, line 88: This "isProducerConsumer" argument should be completely removed from this function. Having an API that accepts an argument, that directly impacts the "how" this API works, instead of the "what" it is supposed to do, means that this isn't a good design for the API.
I suggest you discuss this with @RobiNino and decide on the appropriate refactoring.

artifactory/commands/transferfiles/transfer.go
```diff
@@ -136,27 +136,32 @@ func (tdc *TransferFilesCommand) Run() (err error) {
 			tdc.initNewPhase(newPhase, srcUpService, repoSummary)
 			skip, err := newPhase.shouldSkipPhase()
 			if err != nil {
+				tdc.stopTransferAfterErrorOccurred()
```
eyalbe4, line 139: Notice that we're calling `tdc.stopTransferAfterErrorOccurred()` after when every error is caught here.
The code can call `tdc.stopTransferAfterErrorOccurred()` only once, if we implement the following refactoring:
1. Move the entire code of this loop into a new function.
2. After this function returns an error, call `tdc.stopTransferAfterErrorOccurred()` once.

Let's also rename the "phaseI" into "currentPhase".

artifactory/commands/transferfiles/delayedartifactshandler.go
```diff
@@ -167,18 +172,56 @@ func getDelayUploadComparisonFunctions(packageType string) []shouldDelayUpload {
...
+}
+
+// DelayedArtifactsChannelMng managing writing 'delayed artifacts' to a common channel.
+// In case that an error occurred while handling the files - stops adding elements to the channel.
+type DelayedArtifactsChannelMng struct {
+	channel chan FileRepresentation
+	err     error
+}
+
+// Check if a new element can be added to the channel
```
eyalbe4, line 203: This description does not accurately describe this function, because this function does more than just check,

artifactory/commands/transferfiles/errorshandler.go
```diff
@@ -317,3 +322,35 @@ func getErrorsFiles(repoKey string, isRetry bool) (filesPaths []string, err erro
...
 }
+
+// ErrorsChannelMng managing writing 'files uploading errors' to a common channel.
+// In case that an error occurred while handling the files - stops adding elements to the channel.
+type ErrorsChannelMng struct {
+	channel chan FileUploadStatusResponse
+	err     error
+}
+
+// Check if a new element can be added to the channel
```
eyalbe4, line 333: This description does not accurately describe this function, because this function does more than just check,

artifactory/commands/transferfiles/utils.go
```diff
@@ -142,52 +146,66 @@ func removeTokenFromBatch(uuidTokens []string, token string) []string {
...
+			if !errorsChannelMng.add(file) {
+				log.Debug("Stop transferring data - error occurred while handling transfer's errors files.")
+				return false
+			}
 		case SkippedLargeProps:
-			errorChannel <- file
+			if !errorsChannelMng.add(file) {
+				log.Debug("Stop transferring data - error occurred while handling transfer's errors files.")
+				return false
+			}
```
eyalbe4, line 164: The above lines 156-164 include the same code twice.

eyalbe4 (changes requested):
Since this PR includes many issues that need to be solved, some of them are fundamental, I'd be happy to review it again once all fixed and improved.

artifactory/commands/transferfiles/errorshandler.go
```diff
@@ -317,3 +322,35 @@ func getErrorsFiles(repoKey string, isRetry bool) (filesPaths []string, err erro
...
+}
+
+func (mng ErrorsChannelMng) shouldStop() bool {
+	// Stop adding elements to the channel if an 'blocking' error occurred in a different go routine.
+	return mng.err != nil
+}
+
+func createErrorsChannelMng() ErrorsChannelMng {
+	errorChannel := make(chan FileUploadStatusResponse, fileWritersChannelSize)
+	var writingErrorsErr error
```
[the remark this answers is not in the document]
gailazar300: The declaration of 'writingErrorsErr' is redundant cause like you said it is nil.
During the transfer, the value of the 'err' data member in struct 'ErrorsChannelMng' could be changed from nil to the relevant error value when it occurred.
I removed the declaration of 'writingErrorsErr'.

artifactory/commands/transferfiles/manager.go
```diff
@@ -125,6 +119,43 @@ func (ftm *transferManager) doTransfer(isProducerConsumer bool, transferAction t
...
+	runWaitGroup.Add(1)
+	go func() {
+		defer runWaitGroup.Done()
+		ptm.pollingErr = pollUploads(srcUpService, uploadTokensChan, ptm.doneChannel, errorsChannelMng)
+	}()
+}
+
+func (ptm *PollingTasksManager) stop() {
+	// Notify the other go routines that work is done.
+	ptm.doneChannel <- true
```
eyalbe4, line 150: We can add to the struct an int member variable named `totalRunningGoRoutines`. The variable will be incremented by 1 before each of the two go routines as follows - 
```
	runWaitGroup.Add(1)
        ptm.totalRunningGoRoutines++
	go func() {
		defer runWaitGroup.Done()
		periodicallyUpdateThreads(producerConsumer, ptm.doneChannel)
	}()

	// Check status of uploaded chunks.
	runWaitGroup.Add(1)
        ptm.totalRunningGoRoutines++
	go func() {
		defer runWaitGroup.Done()
		ptm.pollingErr = pollUploads(srcUpService, uploadTokensChan, ptm.doneChannel, errorsChannelMng)
	}()
```

and then, instead of sending 2 close message `stop()` we can use the new `totalRunningGoRoutines` to determine how many messages to send. This way, when we add more polling go routines, we wouldn't need to remember to add another close message.

In addition, notice that you're holding the number 2 here as well - 
```
func newPollingTasksManager() PollingTasksManager {
	// Channel's size is 2, because there are exactly two go routines that need to be signaled that they should stop by it.
	return PollingTasksManager{doneChannel: make(chan bool, 2)}
}
```
The number 2 here can also be used as a struct member. You can call it `totaloGoRoutines`.Now to avoid developers forgetting to update this number, you can compare `totaloRunningGoRoutines` to `totaloGoRoutines` after all go routines are created, and return an error with a clear error message if the number aren't identical.

artifactory/commands/transferfiles/transfer.go
```diff
@@ -233,3 +220,30 @@ type producerConsumerDetails struct {
...
+	// Quit progress bar (before printing logs)
+	if tdc.progressbar != nil {
+		e := tdc.progressbar.Quit()
+		if err == nil {
+			err = e
+		}
+	}
+	// Transferring finished successfully
+	if originalErr == nil {
+		log.Info("Transferring was completed!")
```
eyalbe4, line 238: Transferring was completed! --> Files transfer is complete!

eyalbe4 (changes requested):
Let's add unit tests for the entire error writing functionality. We can do it in a follow up PR, so that @yahavi's work on implementing the graceful shutdown isn't delayed.

artifactory/commands/transferfiles/errorshandler.go
```diff
@@ -317,3 +322,33 @@ func getErrorsFiles(repoKey string, isRetry bool) (filesPaths []string, err erro
 	}
 	return
 }
+
+// ErrorsChannelMng handles the uploading errors and add them to a common channel.
```
eyalbe4, line 326: ...errors and add them... --> ...errors adds them...

artifactory/commands/transferfiles/errorshandler.go
```diff
@@ -132,23 +107,53 @@ func makeDirIfDoesNotExists(path string) error {
...
-func (mng *TransferErrorsMng) start() error {
-	for e := range mng.errorsChannel {
-		err := mng.writeErrorContent(e)
-		if err != nil {
-			return err
-		}
+func (mng *TransferErrorsMng) start() (err error) {
+	// Init content writers manager
+	writerMng := errorWriterMng{}
+	// Init the content writer which responsible for writing 'retryable errors' into files.
```
eyalbe4, line 113: ...writer which responsible... --> ...writer which is responsible...

eyalbe4 (changes requested):
At the bottom of the `initTransferErrorsDir`, the following code - 
```
	err = makeDirIfDoesNotExists(skipped)
	if err != nil {
		return err
	}
	return nil
```
can be simplified to - 
```
	return makeDirIfDoesNotExists(skipped)
```

PR #463, opened 2022-07-19
Config transfer - RB repositories are not filtered

yahavi:
-----
    
Allow filtering release bundle repositories during transfer-config.


artifactory/utils/repositoryutils.go
```diff
@@ -16,13 +17,15 @@ const (
 	REMOTE
 	VIRTUAL
 	FEDERATED
+	RELEASE_BUNDLES
```
eyalbe4, line 20: Go's conversion for consts is came-case and not snake case. Let's convert all the above consts.

eyalbe4 (changes requested):

PR #464, opened 2022-07-19
Config transfer - fail if errors found in the logs

yahavi:
-----
    
* Fail the transfer-config command if `[ERROR]` exists in the import logs.
* Add a hint to use the `--exclude-repos` flag.

Example:
image

### DIFF
--- a/artifactory/commands/transferconfig/transferconfig.go
+++ b/artifactory/commands/transferconfig/transferconfig.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/jfrog/gofrog/version"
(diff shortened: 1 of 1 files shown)


eyalbe4 (changes requested):
@yahavi -
1. The hint about the option of excluding a a specific repository is prompted for any error that is detected in the log. I'm not sure this is something that we'd like to show in any case of a failure. 
2. I'm actually not sure whether this hint is something that we'd like to show. What is the use case in which this will be helpful? Why would I suddenly decide to skip a repository, if it failed to be transferred? I need this repository to be transferred.
3. I suggest we keep the error detection mechanism. It is a good thing in terms of user experience to end the log with an error. We just need a more generic message. How about - 
```
Errors detected during config import. See the above error messages.
```

artifactory/commands/transferconfig/transferconfig.go
```diff
@@ -353,6 +354,9 @@ func (tcc *TransferConfigCommand) waitForImportCompletion(targetServicesManager
 		return err
 	}
 	log.Info(fmt.Sprintf("Logs from Artifactory:\n%s", body))
+	if strings.Contains(string(body), "[ERROR]") {
+		return errorutils.CheckErrorf("Errors detected during config import. Hint: To skip failed repositories, please refer to the '--exclude-repos' flag.")
```
eyalbe4, line 358: Errors detected during config import. Hint: To skip failed repositories, please refer to the '--exclude-repos' flag.
-->
Errors detected during config import. Hint: You can skip transferring some Artifactory repositories by using the '--exclude-repos' command option. Run 'jf rt transfer-config -h' for more information.

eyalbe4 (changes requested):

PR #467, opened 2022-07-24
Add timestamp to upload errors files

gailazar300:
-----


artifactory/commands/transferfiles/filesdiff.go
```diff
@@ -281,7 +281,7 @@ func (f *filesDiffPhase) handleErrorsFiles(uploadTokensChan chan string, delayHe
 	return nil
 }
 
-func convertUploadStatusToFileRepresentation(statuses []FileUploadStatusResponse) (files []FileRepresentation) {
+func convertUploadStatusToFileRepresentation(statuses []ExtendedFileUploadStatusResponse) (files []FileRepresentation) {
```
RobiNino, line 284: We need to take into account that users will not be able to consume errors files that were created before this change. Maybe a breaking change comment in the release notes is enough.

RobiNino (changes requested):

artifactory/commands/transferfiles/filesdiff.go
gailazar300: The convert function care only about the 'FileRepresentation' - so changing 'FileUploadStatusResponse' to 'ExtendedFileUploadStatusResponse' shouldn't break anything, just adding more information to the summary file. (also tested it by hand to saw that it doesn't file with old files structure)


```
func convertUploadStatusToFileRepresentation(statuses []ExtendedFileUploadStatusResponse) (files []FileRepresentation) {
	for _, status := range statuses {
		files = append(files, status.FileRepresentation)
	}
	return
}
```

PR #475, opened 2022-07-31
Transfer - remove chunk upload checksum deploy support

RobiNino:
-----


eyalbe4 (approved):

utils/transfer/utils.go
```diff
@@ -0,0 +1,25 @@
...
+// t           - The testing object
+// testHandler - The HTTP handler of the test
+func CreateRestsMockServer(t *testing.T, testHandler restsTestHandler) (*httptest.Server, *config.ServerDetails, artifactory.ArtifactoryServicesManager) {
+	testServer := httptest.NewServer(http.HandlerFunc(testHandler))
+	serverDetails := &config.ServerDetails{ArtifactoryUrl: testServer.URL + "/"}
+
+	serviceManager, err := utils.CreateServiceManager(serverDetails, -1, 0, false)
+	assert.NoError(t, err)
+	return testServer, serverDetails, serviceManager
+}
```
yahavi, line 25: Since this is a generic component, I suggest moving the content to utils/tests/utils.go

artifactory/commands/transferfiles/transfer_test.go
```diff
@@ -0,0 +1,162 @@
...
+	for _, test := range testsArray {
+		t.Run(test.testName, func(t *testing.T) {
+			pluginVersion = test.curVersion
+			err = getAndValidateDataTransferPlugin(srcPluginManager)
+			if test.errorExpected {
+				assert.EqualError(t, err, getMinimalVersionErrorMsg(test.curVersion))
+				return
+			} else {
+				assert.NoError(t, err)
+			}
```
yahavi, line 56:
```suggestion
			}
			assert.NoError(t, err)
```

artifactory/commands/transferfiles/transfer_test.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		assert.NoError(t, err)
+		return
+	}
+
+	_, err = srcPluginManager.version()
+	assert.ErrorContains(t, err, "Response from Artifactory: 404 Not Found.")
+}
+
+func TestUploadChunkAndPollUploads(t *testing.T) {
```
yahavi, line 80: This function is way too big and 
Let's break it into more functions to make it more readable.

artifactory/commands/transferfiles/transfer_test.go
```diff
@@ -0,0 +1,162 @@
...
+			}
+		})
+	}
+}
+
+func TestVerifyConfigImportPluginNotInstalled(t *testing.T) {
+	testServer, serverDetails, _ := transfer.CreateRestsMockServer(t, func(w http.ResponseWriter, r *http.Request) {
+		if r.RequestURI == "/"+pluginsExecuteRestApi+"dataTransferVersion" {
+			w.WriteHeader(http.StatusNotFound)
+			_, _ = w.Write([]byte("Not found"))
```
yahavi, line 65:
```suggestion
			_, err := w.Write([]byte("Not found"))
			assert.NoError(t, err)
```

artifactory/commands/transferfiles/transfer_test.go
```diff
@@ -0,0 +1,162 @@
...
+	})
+	defer testServer.Close()
+
+	srcPluginManager, err := createSrcRtUserPluginServiceManager(serverDetails)
+	if err != nil {
+		assert.NoError(t, err)
+		return
+	}
+
+	testsArray := []validateVersionTestSuite{
```
yahavi, line 41: I think the convention in Table Testing is to put the test cases outside the test method. This way it makes the method shorter and more readable.
Let's make this change in the other tests too.

artifactory/commands/transferfiles/transfer_test.go
```diff
@@ -0,0 +1,162 @@
...
+		if r.RequestURI == "/"+pluginsExecuteRestApi+"dataTransferVersion" {
+			content, err := json.Marshal(utils.VersionResponse{Version: pluginVersion})
+			assert.NoError(t, err)
+			_, err = w.Write(content)
+			assert.NoError(t, err)
+		}
+	})
+	defer testServer.Close()
+
+	srcPluginManager, err := createSrcRtUserPluginServiceManager(serverDetails)
```
yahavi, line 35: This code is repeated for each test method. Let's give the tests equal respect as the production code.
I suggest exporting it to a function. (we don't really need the "return", because we always look at the first assertion anyway)

artifactory/commands/transferfiles/transfer_test.go
```diff
@@ -0,0 +1,162 @@
...
+	uploadTokensChan := make(chan string, 3)
+	doneChan := make(chan bool, 1)
+
+	// Sends chunk to upload, polls on chunk twice - once when it is still in progress, and once after it is done.
+	err = uploadChunkAndAddToken(srcPluginManager, UploadChunk{
+		UploadCandidates: []FileRepresentation{fileSample},
+	}, uploadTokensChan)
+	if err != nil {
+		assert.NoError(t, err)
+		return
```
yahavi, line 144:
```suggestion
```

artifactory/commands/transferfiles/srcpluginservice.go
```diff
@@ -75,19 +74,18 @@ func (sup *srcUserPluginService) uploadChunk(chunk UploadChunk) (uuidToken strin
...
-		return "", nil
-	}
-
 	var uploadResponse UploadChunkResponse
 	err = json.Unmarshal(body, &uploadResponse)
 	if err != nil {
 		return "", errorutils.CheckError(err)
 	}
+	if uploadResponse.UuidToken == "" {
+		return "", errorutils.CheckErrorf("unexpected empty token return for chunk upload")
```
yahavi, line 87:
```suggestion
		return "", errorutils.CheckErrorf("unexpected empty token returned for chunk upload")
```

yahavi (changes requested):
Looks good!
Please consider my comments.

artifactory/commands/transferfiles/transfer_test.go
RobiNino: It is on purpose so we will fail fast

utils/transfer/utils.go
RobiNino: Cannot mode there due to cyclic import, made it clear that it is used for tests

PR #478, opened 2022-08-01
Lock plugins config file only if needed

sverdlov93:
-----


utils/plugins/utils.go
```diff
@@ -70,23 +51,52 @@ func CheckPluginsVersionAndConvertIfNeeded() (err error) {
...
+		defer func() {
+			e := unlockFunc()
+			if err == nil {
+				err = e
+			}
+		}()
+		if err != nil {
+			return
+		}
+		// The reason behind reading the config again is that its possible that the plugins file already changed by other thread,
```
talarian1, line 81:
```suggestion
		// The reason behind reading the config again is that it's possible that another thread or process already changed the plugins file,
```

utils/plugins/utils.go
```diff
@@ -60,33 +40,49 @@ func CheckPluginsVersionAndConvertIfNeeded() (err error) {
...
-	}
-	return
+	return readPluginsConfigAndConvertV0tToV1IfNeeded()
 }
 
-func readPluginsConfig() (*PluginsV1, error) {
-	plugins := new(PluginsV1)
+func readPluginsConfigAndConvertV0tToV1IfNeeded() (err error) {
 	content, err := getPluginsConfigFileContent()
 	if err != nil {
```
RobiNino, line 48: Can be simplified by checking:
```suggestion
        // Return without converting in case of an error, or if the plugins.yml file already exists, which indicates a conversion has already been made.
	if err != nil || len(content)!=0 {
```
Then removing the second if.

RobiNino (changes requested):
Thanks!

PR #492, opened 2022-08-08
Transfer files ignoreState option

asafgabai:
-----
Added ignoreState flag to TransferFilesCommand. If true, the CLI will ignore the saved state of previous runs of the transfer.

### DIFF
--- a/artifactory/commands/transferfiles/state_test.go
+++ b/artifactory/commands/transferfiles/state_test.go
@@ -51,7 +51,7 @@ func getAndAssertNonExistingRepo(t *testing.T, state *TransferState, repoKey str
 }
 
 func TestFilesDiffRange(t *testing.T) {
-	cleanUp := setupStateTest(t)
+	cleanUp := initStateTest(t)
 	defer cleanUp()
 
 	repoKey := "repo"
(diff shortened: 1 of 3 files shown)


artifactory/commands/transferfiles/transfer.go
```diff
@@ -152,6 +157,13 @@ func (tdc *TransferFilesCommand) transferRepos(sourceRepos []string, targetRepos
 			tdc.progressbar.NewRepository(repoKey)
 		}
 
+		if tdc.ignoreState {
+			err = resetRepoState(repoKey)
+			if err != nil {
+				return tdc.cleanup(err, sourceRepos)
```
yahavi, line 163: The cleanup already gets triggered when an error is returned from this method
```suggestion
				return err
```

artifactory/commands/transferfiles/state_test.go
```diff
@@ -129,3 +122,38 @@ func getRepoFromState(t *testing.T, repoKey string) *Repository {
 	assert.NoError(t, err)
 	return repo
 }
+
+func setupStateTest(t *testing.T) (cleanUp func()) {
```
yahavi, line 126: Our convention is to call this function init:
```suggestion
func initStateTest(t *testing.T) (cleanUp func()) {
```

yahavi (changes requested):

PR #496, opened 2022-08-09
Transfer - add repository state manager

RobiNino:
-----


eyalbe4 (changes requested):
* We need to undertand why we're calling the `IncrementPhase` in the `handleErrorsFiles` functions. Since this action is part of *phase 2*, for which the progress shows time slots, how come we increase the progress for every error file?
* Can we actually remove the `IncrementPhase` alltogether, and just use `IncrementPhaseBy` with `1` as the increased value?

RobiNino:
@eyalbe4 ,
Regarding `IncrementPhase` in the `handleErrorsFiles` - the correct total of the progress bar is set in `initProgressBar`, and includes the number of errors files. I think it makes sense to include it in the total and increment as we do know, to show work is being done in every stage of the phase.

utils/repostate/statemanager.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	relativePath = strings.TrimSuffix(relativePath, "/")
+	if relativePath == "." {
+		return sm.Root, nil
+	}
+
+	// Progress through the children maps till reaching the node that represents the requested path.
+	dirs := strings.Split(relativePath, "/")
+	curNode := sm.Root
+dirsLoop:
```
eyalbe4, line 94: I tried going through this code, and couldn't understand the logic of these loops. There's a good chance the code does what it is supposed to, but let's see if we can make it easier to read and understanbd.
I'd be happy to review it again your changes.

utils/repostate/statemanager.go
```diff
@@ -0,0 +1,194 @@
...
+func (sm *RepoStateManager) SaveToFile() error {
+	content, err := json.Marshal(sm)
+	if err != nil {
+		return errorutils.CheckError(err)
+	}
+	return errorutils.CheckError(ioutil.WriteFile(sm.stateFilePath, content, 0644))
+}
+
+// Returns the node corresponding to the directory in the provided relative path. Path should be provided without the repository name.
+func (sm *RepoStateManager) LookUpNode(relativePath string) (*Node, error) {
```
eyalbe4, line 82: Don't we have an issue of thread safety here? If multiple go routine can call this API simultaneously, we should find the best and most efficient way to protect the critical sections of the data structure.
If indeed we have an issue here, and we need to modify the code, I'd be happy to revierw the code again after those changes.

eyalbe4 (changes requested):
I haven't finished going over all the code. I'll continue review it soon.
I'm submitting here what I have so far.

eyalbe4 (changes requested):
Done reviewing everything.
I'd be happy to review it again following the changes.

PR #504, opened 2022-08-14
Transfer - Missing repository in the repositories summary

yahavi:
-----
    
Depends on:
* https://github.com/jfrog/build-info-go/pull/93
* https://github.com/jfrog/jfrog-client-go/pull/633

Currently, we start storage info calculation right before the storage info retrieval. As a result, the storage info may not be updated, and may not contain new repositories.

* Create StorageInfoManager
* Refactor transfer

### DIFF
--- a/artifactory/commands/transferfiles/transfer.go
+++ b/artifactory/commands/transferfiles/transfer.go
@@ -112,12 +112,23 @@ func (tdc *TransferFilesCommand) Run() (err error) {
 }
 
 func (tdc *TransferFilesCommand) initStorageInfoManagers() error {
-	tdc.sourceStorageInfoManager = utils.NewStorageInfoManager(tdc.sourceServerDetails)
-	if err := tdc.sourceStorageInfoManager.CalculateStorageInfo(); err != nil {
+	// Init source storage info manager
+	storageInfoManager, err := utils.NewStorageInfoManager(tdc.sourceServerDetails)
+	if err != nil {
 		return err
 	}
-	tdc.targetStorageInfoManager = utils.NewStorageInfoManager(tdc.targetServerDetails)
-	return tdc.targetStorageInfoManager.CalculateStorageInfo()
+	tdc.sourceStorageInfoManager = storageInfoManager
+	if err := storageInfoManager.CalculateStorageInfo(); err != nil {
+		return err
+	}
+
+	// Init target storage info manager
+	storageInfoManager, err = utils.NewStorageInfoManager(tdc.targetServerDetails)
+	if err != nil {
+		return err
+	}
+	tdc.targetStorageInfoManager = storageInfoManager
+	return storageInfoManager.CalculateStorageInfo()
 }
 
 func (tdc *TransferFilesCommand) transferRepos(sourceRepos []string, targetRepos []string,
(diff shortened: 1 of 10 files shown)


artifactory/utils/storageinfo.go
```diff
@@ -0,0 +1,73 @@
...
+var getRepoSummaryPollingInterval = 10 * time.Second
+
+type StorageInfoManager struct {
+	serviceManager artifactory.ArtifactoryServicesManager
+}
+
+func NewStorageInfoManager(serverDetails *config.ServerDetails) *StorageInfoManager {
+	serviceManager, err := CreateServiceManager(serverDetails, serviceManagerRetriesPerRequest, serviceManagerRetriesWaitPerRequest, false)
+	if err != nil {
+		return nil
```
eyalbe4, line 29: We should return an error here.

artifactory/utils/storageinfo.go
```diff
@@ -0,0 +1,73 @@
...
+	return sim.serviceManager.CalculateStorageInfo()
+}
+
+// Get storage info from Artifactory
+func (sim *StorageInfoManager) GetStorageInfo() (*utils.StorageInfo, error) {
+	return sim.serviceManager.GetStorageInfo()
+}
+
+// Get repository summary from the storage info.
+// This method must be running after CalculateStorageInfo.
```
eyalbe4, line 45: ...must be running after... --> ...must be called after...

artifactory/commands/transferfiles/transfer.go
```diff
@@ -68,108 +72,144 @@ func (tdc *TransferFilesCommand) Run() (err error) {
...
+			log.Error("repository '" + repoKey + "' does not exist in target. Skipping...")
 			continue
 		}
 
-		repoSummary, err := getRepoSummaryFromList(sourceStorageInfo.RepositoriesSummaryList, repo)
+		repoSummary, err := tdc.sourceStorageInfoManager.GetRepoSummary(repoKey)
 		if err != nil {
 			log.Error(err.Error() + ". Skipping...")
-			continue
+			return nil
```
eyalbe4, line 137: I think we should continue and not return nil. This will actually skip repo (as the above log message says). This can happen in case the repo was deleted from the source, before the transfer process reached it.

artifactory/utils/storageinfo.go
```diff
@@ -0,0 +1,73 @@
...
+					retVal = &repoSummary
+					return true, []byte{}, nil
+				}
+			}
+			return false, []byte{}, nil
+		},
+	}
+	_, err := pollingExecutor.Execute()
+	if retVal == nil && err == nil {
+		return nil, errorutils.CheckErrorf("could not find repository '%s' in the repositories summary", repoKey)
```
eyalbe4, line 70: Let;s make this error message a const, so that the test can also use it, instead fo duplicating it.

eyalbe4 (changes requested):

eyalbe4 (changes requested):
Also notice the static code analysis errors.

artifactory/utils/storageinfo.go
yahavi: I removed the other usage of this message.

PR #505, opened 2022-08-15
Handle typo in access.bootstrap.json

asafgabai:
-----


artifactory/commands/transferconfig/utils.go
```diff
@@ -37,6 +38,23 @@ func archiveConfig(exportPath string, configXml string) (buffer *bytes.Buffer, r
...
+	if !accessBootstrapExists {
+		err = fileutils.MoveFile(filepath.Join(exportPath, "etc", "access.boostrap.json"), accessBootstrapPath)
+		if err != nil {
+			if os.IsNotExist(err) {
+				return nil, errorutils.CheckErrorf("%s: the file was not found or is not accessible", accessBootstrapPath)
+			}
+			return nil, err
+		}
+	}
+
```
yahavi, line 57: I think this logic would look better in a new function.
This new function can have a nice unit test.

artifactory/commands/transferconfig/utils_test.go
```diff
@@ -35,3 +36,17 @@ func TestArchiveConfig(t *testing.T) {
 		}
 	}
 }
+
+// In some versions of Artifactory, the file access.bootstrap.json has a typo in its name: access.boostrap.json.
+// This tests this scenario.
+func TestArchiveConfigWithTypoInAccessBootstrap(t *testing.T) {
```
yahavi, line 42: I think that instead of duplicating artifactory_export, maybe it would be better to export TestArchiveConfig content to an internal function testArchiveConfig(t *testing.T).
After that, we can run 2 tests:
1. `TestArchiveConfig` - Just run testArchiveConfig(t)
2. `TestArchiveConfigBoostrap` - Rename access.bootstap.json -> access.boostrap.json and run testArchiveConfig(t)

yahavi (changes requested):
Looks good!

What do you think about moving `access.boostrap.json` -> `access.bootsrap.json`, to add `access.boostrap.json` directly to the archive? This way we will reduce redundant I/Os and a redundant check that  `access.bootstrap.json` is exists.

asafgabai:
> Looks good!
> 
> What do you think about moving `access.boostrap.json` -> `access.bootsrap.json`, to add `access.boostrap.json` directly to the archive? This way we will reduce redundant I/Os and a redundant check that `access.bootstrap.json` is exists.

I considered doing it, but this scenario is very rare and the check if the file exists is pretty cheap in terms of performance.
On the other hand, your suggestion requires adding a check inside the loop, making it less readable and less understandable.

PR #516, opened 2022-08-24
Avoid old snapshots/tags cleanup after transfer

asafgabai:
-----
Turn off the Max Unique Snapshots/Tags setting at the beginning of the transfer-files operation and turn it back on at the end.
There might be more snapshots/tags than the maximum if the Max Unique Snapshots/Tags property was set and no snapshot/tag was uploaded afterward.
In cases like this, we don't want to trigger the cleanup of old snapshots, so we turn off the Max Unique Snapshots/Tags setting during the transfer.
This setting exists only in repositories of these package types: Maven, Gradle, NuGet, Ivy, SBT, and Docker.


artifactory/commands/transferfiles/transfer.go
```diff
@@ -164,6 +164,20 @@ func (tdc *TransferFilesCommand) transferRepos(sourceRepos []string, targetRepos
...
 		}
 
+		// Get the source repository's max unique snapshots setting
+		srcMaxUniqueSnapshots, err := getMaxUniqueSnapshots(tdc.sourceServerDetails, repoSummary)
+		if err != nil {
+			return err
+		}
+
+		// If it's a Maven, Gradle, NuGet, Ivy, SBT or Docker repository, update its max unique snapshots setting to 0.
+		if srcMaxUniqueSnapshots != -1 {
```
omerzi, line 174: Maybe let's add a comment about the -1 meaning?

artifactory/commands/transferfiles/utils.go
```diff
@@ -417,3 +418,120 @@ func stopTransferOnArtifactoryNodes(srcUpService *srcUserPluginService, runningN
...
+// or max unique tags (Docker).
+// On repositories of other package types or if an error is thrown, this function returns -1.
+func getMaxUniqueSnapshots(rtDetails *coreConfig.ServerDetails, repoSummary *serviceUtils.RepositorySummary) (maxUniqueSnapshots int, err error) {
+	maxUniqueSnapshots = -1
+	serviceManager, err := utils.CreateServiceManager(rtDetails, retries, retriesWaitMilliSecs, false)
+	if err != nil {
+		return
+	}
+	switch repoSummary.PackageType {
+	case maven:
```
omerzi, line 432: I think we can use the Factory Method design pattern in order to create concrete LocalRepositoryParam, and then we can reduce the number of duplicated lines, this also applies to 'updateMaxUniqueSnapshots'

artifactory/commands/transferfiles/utils.go
```diff
@@ -417,3 +418,120 @@ func stopTransferOnArtifactoryNodes(srcUpService *srcUserPluginService, runningN
...
+// updateMaxUniqueSnapshots updates the local repository's setting of max unique snapshots (Maven, Gradle, NuGet, Ivy and SBT)
+// or max unique tags (Docker).
+// On repositories of other package types, this function does nothing.
+func updateMaxUniqueSnapshots(rtDetails *coreConfig.ServerDetails, repoSummary *serviceUtils.RepositorySummary, newMaxUniqueSnapshots int) error {
+	serviceManager, err := utils.CreateServiceManager(rtDetails, retries, retriesWaitMilliSecs, false)
+	if err != nil {
+		return err
+	}
+	switch repoSummary.PackageType {
+	case maven:
```
omerzi, line 487: See comment above about the factory method

omerzi (changes requested):
Seems good to me overall, I think we can reduce some duplicated code using the Factory method, contact me if you need further explanation on what I mean

artifactory/commands/transferfiles/transfer.go
```diff
@@ -177,6 +192,14 @@ func (tdc *TransferFilesCommand) transferRepos(sourceRepos []string, targetRepos
...
 			}
 		}
+
+		// Update the target repository's max unique snapshots setting to be the same as in the source, only if it's not 0.
+		if srcMaxUniqueSnapshots > 0 {
+			err = updateMaxUniqueSnapshots(tdc.targetServerDetails, repoSummary, srcMaxUniqueSnapshots)
+			if err != nil {
+				return err
+			}
+		}
```
eyalbe4, line 202: I think that this code should be put inside a defer clause, right after calling the new `handleMaxUniqueSnapshots` as I suggested above. Actually, this code can be inside a function named `restoreTargetMaxUniqueSnapshots`. This function can be a returned value of the `handleMaxUniqueSnapshots` function.

artifactory/commands/transferfiles/utils.go
```diff
@@ -417,3 +418,120 @@ func stopTransferOnArtifactoryNodes(srcUpService *srcUserPluginService, runningN
 		}
 	}
 }
+
+// getMaxUniqueSnapshots gets the local repository's setting of max unique snapshots (Maven, Gradle, NuGet, Ivy and SBT)
+// or max unique tags (Docker).
+// On repositories of other package types or if an error is thrown, this function returns -1.
```
eyalbe4, line 424: On repositories... --> For repositories...

artifactory/commands/transferfiles/utils.go
```diff
@@ -417,3 +418,120 @@ func stopTransferOnArtifactoryNodes(srcUpService *srcUserPluginService, runningN
...
+			return
+		}
+		maxUniqueSnapshots = *dockerLocalRepoParams.MaxUniqueTags
+	}
+	return
+}
+
+// updateMaxUniqueSnapshots updates the local repository's setting of max unique snapshots (Maven, Gradle, NuGet, Ivy and SBT)
+// or max unique tags (Docker).
+// On repositories of other package types, this function does nothing.
```
eyalbe4, line 480: On repositories... --> For repositories...

eyalbe4 (changes requested):
I think this PR adds the first API of the `transfer-files` command, which invokes an API on the target instance directly, and not through the `data-transfer` user plugin.
I therefore think we should try pinging the target instance at the very beginning of the transfer process, and fail the command with a clear error message, in case the target is not accessible.

asafgabai:
> I think this PR adds the first API of the `transfer-files` command, which invokes an API on the target instance directly, and not through the `data-transfer` user plugin. I therefore think we should try pinging the target instance at the very beginning of the transfer process, and fail the command with a clear error message, in case the target is not accessible.

We get the repositories list from the target beforehand, so if it's not accessible, it will fail there.

PR #524, opened 2022-08-28
Warn when configuring unsafe HTTP connections

yahavi:
-----
    
Warn when a user configures an HTTP connection to the platform:
Interactive:
image

Non-interactive:
image

### DIFF
--- a/common/commands/config.go
+++ b/common/commands/config.go
@@ -640,11 +640,11 @@ func (cc *ConfigCommand) assertUrlsSafe() error {
 			continue
 		}
 		if cc.interactive {
-			if cc.disablePrompts || !coreutils.AskYesNo("We noticed that your JFrog URL uses an insecure HTTP connection. Are you sure you want to continue?", false) {
+			if cc.disablePrompts || !coreutils.AskYesNo("Your JFrog URL uses an insecure HTTP connection, instead of HTTPS. Are you sure you want to continue?", false) {
 				return errorutils.CheckErrorf("config was aborted due to an insecure HTTP connection")
 			}
 		} else {
-			log.Warn("Your configured JFrog URL uses an insecure HTTP connection. Please consider replacing it with HTTPS.")
+			log.Warn("Your configured JFrog URL uses an insecure HTTP connection. Please consider using SSL (HTTPS instead of HTTP).")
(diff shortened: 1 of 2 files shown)


common/commands/config.go
```diff
@@ -199,6 +203,17 @@ func (cc *ConfigCommand) config() error {
 	return config.SaveServersConf(configurations)
 }
 
+func (cc *ConfigCommand) isUrlUnsafe() bool {
+	artifactoryUrl := cc.details.Url
+	if !strings.HasPrefix(artifactoryUrl, "http://") {
+		return false
+	}
+	if strings.Contains(artifactoryUrl, "127.0.0.1") || strings.Contains(artifactoryUrl, "localhost") {
```
eyalbe4, line 211: I think that `strings.contains` isn't accurate enough for this check. For example, I can use a domain name that contains "localhost" but isn't "localhost", for example "localhost-123" and this test will pass. I therefore suggest to split the URL using a good utility, and then using `equals` to test this.

eyalbe4 (changes requested):

common/commands/config.go
```diff
@@ -626,6 +632,47 @@ func (cc *ConfigCommand) encryptPassword() error {
...
 
+// Assert all services URLs are safe
+func (cc *ConfigCommand) assertUrlsSafe() error {
+	for _, url := range []string{cc.details.Url, cc.details.AccessUrl, cc.details.ArtifactoryUrl,
+		cc.details.DistributionUrl, cc.details.MissionControlUrl, cc.details.PipelinesUrl, cc.details.XrayUrl} {
+		if isUrlSafe(url) {
+			continue
+		}
+		if cc.interactive {
+			if cc.disablePrompts || !coreutils.AskYesNo("We noticed that your JFrog URL uses an insecure HTTP connection. Are you sure you want to continue?", false) {
```
eyalbe4, line 643: We noticed that your JFrog URL uses an insecure HTTP connection. Are you sure you want to continue?
-->
Your JFrog URL uses an insecure HTTP connection, instead of HTTPS. Are you sure you want to continue?

eyalbe4 (changes requested):

PR #535, opened 2022-09-04
Transfer files - better handling graceful stop

yahavi:
-----
    
Depends on https://github.com/jfrog/jfrog-client-go/pull/647

Fix the following issues:
1. Create a cancellable context and send it to the Go client, to allow retry executables to be canceled
2. Add "ShouldStop" before running the directory AQL
3. Stop the active producer-consumers

### DIFF
--- a/artifactory/commands/transferfiles/fulltransfer.go
+++ b/artifactory/commands/transferfiles/fulltransfer.go
@@ -18,6 +18,7 @@ import (
 // New folders found are handled as a separate task, and files are uploaded in chunks and polled on for status.
 type fullTransferPhase struct {
 	phaseBase
+	transferManager *transferManager
 }
 
 func (m *fullTransferPhase) getSourceDetails() *coreConfig.ServerDetails {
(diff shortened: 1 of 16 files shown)


artifactory/commands/transferfiles/fulltransfer.go
```diff
@@ -58,10 +59,13 @@ func (m *fullTransferPhase) phaseStarted() error {
 }
 
 func (m *fullTransferPhase) phaseDone() error {
-	err := setRepoFullTransferCompleted(m.repoKey)
-	if err != nil {
-		return err
+	// If the phase stopped gracefully (with ctrl+c), don't mark the phase as completed
```
eyalbe4, line 62: Since I think we should also add a `--stop` option for the `transfer-files` command (for cases when the command is ran a background process) I suggest removing the `(with ctrl+c)` from thre comment, since it'll soon not be the only method to stop the process.

The same goes for `filesdiff.go`.

artifactory/commands/transferfiles/utils.go
```diff
@@ -350,7 +340,7 @@ func uploadChunkWhenPossibleHandler(phaseBase *phaseBase, chunk UploadChunk, upl
 		shouldStop := uploadChunkWhenPossible(phaseBase, chunk, uploadTokensChan, errorsChannelMng)
 		if shouldStop {
 			// The specific error that triggered the stop is already in the errors channel
-			return errorutils.CheckErrorf("%s stopped.", logMsgPrefix)
+			return errorutils.CheckErrorf("%sstopped.", logMsgPrefix)
```
eyalbe4, line 343: Unless this white-space was removed intentionally, let's add it back.

eyalbe4 (changes requested):

artifactory/commands/transferfiles/utils.go
yahavi: This is intentional - there is already a space in the log prefix:
https://github.com/jfrog/jfrog-client-go/blob/v1.22.0/utils/utils.go#L309

PR #545, opened 2022-09-12
Fix Dotnet configfile usage

RobiNino:
-----
    
 Following https://github.com/jfrog/jfrog-cli/issues/1623.
The configfile flag was not passed correctly to the Dotnet module when it was generated.
Breaking change: Removed unused setter `SetUseNugetAddSource`
Depends on https://github.com/jfrog/build-info-go/pull/103.

### DIFF
--- a/artifactory/commands/dotnet/dotnetcommand.go
+++ b/artifactory/commands/dotnet/dotnetcommand.go
@@ -18,7 +18,10 @@ import (
 	"strings"
 )
 
-const SourceName = "JFrogCli"
+const (
+	SourceName        = "JFrogCli"
+	configFilePattern = "jfrog.cli.nuget."
+)
 
 type DotnetCommand struct {
 	toolchainType      dotnet.ToolchainType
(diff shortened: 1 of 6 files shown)


artifactory/commands/dotnet/dotnetcommand.go
```diff
@@ -18,19 +18,20 @@ import (
 	"strings"
 )
 
-const SourceName = "JFrogCli"
+const (
+	SourceName        = "JFrogCli"
```
eyalbe4, line 22: It looks like this const can be made private.

artifactory/commands/dotnet/dotnetcorecli.go
```diff
@@ -20,27 +15,5 @@ func NewDotnetCoreCliCommand() *DotnetCoreCliCommand {
 }
 
 func (dccc *DotnetCoreCliCommand) Run() (err error) {
-	dccc.useNugetAddSource, err = isDotnetVersionAboveMin()
```
eyalbe4, line 23: What is the reason the min version check was removed? Was this check move to build-info-go?

eyalbe4 (changes requested):

artifactory/commands/dotnet/dotnetcorecli.go
RobiNino: It was a check for the minimal supported version of the `nuget add source` command. However, the usage of this command was previously removed because it does not support configuring the protocol version.

artifactory/commands/dotnet/dotnetcommand.go
RobiNino: It is used by the tests in the CLI

eyalbe4: Got it. NP.

artifactory/commands/dotnet/dotnetcorecli.go
eyalbe4: Got it. Thanks.

PR #548, opened 2022-09-12
Transfer files - remaining time estimation

asafgabai:
-----
Added estimation of the remaining time in transfer-files command.
Tests will be added in a separate PR.


artifactory/utils/storageinfo.go
```diff
@@ -71,3 +73,78 @@ func (sim *StorageInfoManager) GetRepoSummary(repoKey string) (*utils.Repository
...
+			reposMap := make(map[string]bool)
+			for _, repoKey := range repoKeys {
+				reposMap[repoKey] = true
+			}
+			for _, repoSummary := range storageInfo.RepositoriesSummaryList {
+				if reposMap[repoSummary.RepoKey] {
+					reposCounted++
+					if repoSummary.UsedSpaceInBytes.String() != "" {
+						sizeToAdd, err := repoSummary.UsedSpaceInBytes.Int64()
+						if err != nil {
```
eyalbe4, line 101: Not sure I understand this line - 
If we received an error, we append the value and continue? What is the reason for this?

artifactory/utils/storageinfo.go
```diff
@@ -71,3 +73,78 @@ func (sim *StorageInfoManager) GetRepoSummary(repoKey string) (*utils.Repository
...
+}
+
+func convertStorageSizeStringToBytes(sizeStr string) (int64, error) {
+	usedSpaceParts := strings.Split(sizeStr, " ")
+	if len(usedSpaceParts) != 2 {
+		return 0, errorutils.CheckErrorf("could not parse size string '%s'", sizeStr)
+	}
+	sizeInUnit, err := strconv.ParseFloat(usedSpaceParts[0], 64)
+	if err != nil {
+		return 0, err
```
eyalbe4, line 131: Missing CheckError here.

eyalbe4 (changes requested):

eyalbe4 (changes requested):
Also notice that the Static Analysis test is failing.

eyalbe4 (changes requested):
If spped / time estimation data isn't available yet, let's show 
"Data is not yet available, since only metadata has been transferred till now"

Let's also color the "speed" and "time remaining" values in green, so that it looks compatible with the "current repository" value. We have an API that can color the text in green. 

image

artifactory/commands/transferfiles/timeestimation.go
```diff
@@ -0,0 +1,94 @@
...
+	return remainingTime / 1000
+}
+
+// getEstimatedRemainingTimeString gets the estimated remaining time in an easy-to-read string.
+func (tem *timeEstimationManager) getEstimatedRemainingTimeString() string {
+	remainingTimeSec := tem.getEstimatedRemainingTime()
+	remainingDays := remainingTimeSec / secondsInDay
+	remainingDaysInSecs := remainingDays * secondsInDay
+	remainingHours := (remainingTimeSec - remainingDaysInSecs) / secondsInHour
+	if remainingDays > 0 {
```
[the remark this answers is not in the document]
asafgabai: `remainingDays` is of type int64 so it can't, but I'll take your suggestion anyways to make the code clearer and make more sense.

yahavi:
Since the calculation can't be 100% accurate, can we add the "estimation" word to the "Transfer speed" and "Time remaining"? 
```
Transfer speed estimation:
Estimated time remaining:
```

artifactory/utils/storageinfo.go
asafgabai: My mistake. It should be `if err == nil...`

artifactory/utils/storageinfo.go
```diff
@@ -71,3 +73,78 @@ func (sim *StorageInfoManager) GetRepoSummary(repoKey string) (*utils.Repository
...
+	case "bytes":
+		sizeInBytes = sizeInUnit
+	case "KB":
+		sizeInBytes = sizeInUnit * 1024
+	case "MB":
+		sizeInBytes = sizeInUnit * 1024 * 1024
+	case "GB":
+		sizeInBytes = sizeInUnit * 1024 * 1024 * 1024
+	case "TB":
+		sizeInBytes = sizeInUnit * 1024 * 1024 * 1024 * 1024
```
yahavi, line 145: Let's use consts here to reduce redundant calculations and to be more clear. For example:
```go
const (
    _ = 1 << (10 * iota)
    KiBInBytes
    MiBInBytes
    GiBInBytes
    TiBInBytes
)
```

yahavi (changes requested):

yahavi:
💡  Suggestions:
Instead of: `Time remaining: About 0 hours and 0 minutes`
Show: `Estimated time remaining: Less than a minute` or "not yet available" as @eyalbe4 suggested

Instead of: `Time remaining: About 0 hours and 4 minutes`
Show: `Estimated time remaining: 4 minutes`

PR #581, opened 2022-10-14
Support new transfer status option

yahavi:
-----
    
Requires for https://github.com/jfrog/jfrog-cli/pull/1722

* Support transfer status option
* Create a state-manager containing the old state handling and a new run-status.

Example:
image


eyalbe4 (changes requested):
I haven't reviewed the code yet, but I'm sharing my suggested changes to the displayed labels, following the screenshot added.

1. Change the value of the "Status" line to one of the following - "Running" / "Not running", instead of "Transferring files".

2. Transferred size --> Storage

3. Let's add a line for "Files" (showing total files) before the "Storage" line. It is important in my opinion that the info displayed in the status output matches the data displayed in the terminal display. There's a chance Sara already added the code that calculates it.

4. Let's also add a line with the label "Failures" with the value of the new failures counter recently added by @attiasas. 

5. Transferred repositories --> Repositories

6. Max working threads --> Working threads 
Let's make sure we're showing the total of current working threads and not the max working threads, just like the terinal view displayes.

7. Transferred files --> Files

8. Transferred data --> Storage
Let's put the "Files" line before the "Storage" line.

artifactory/commands/transferfiles/state/statemanager.go
```diff
@@ -0,0 +1,250 @@
...
+		return nil
+	})
+}
+
+func (ts *TransferStateManager) SaveState() error {
+	ts.TransferRunStatus.lastSaveTimestamp = time.Now()
+	return ts.saveTransferState()
+}
+
+func (ts *TransferStateManager) lockStateManager() error {
```
eyalbe4, line 231: Let's add a description for this function.

artifactory/commands/transferfiles/state/statemanager.go
```diff
@@ -0,0 +1,250 @@
...
+}
+
+func (ts *TransferStateManager) GetTransferredSizeBytes(transferredSizeBytes *int64) error {
+	return ts.TransferRunStatus.action(func(transferRunStatus *TransferRunStatus) error {
+		*transferredSizeBytes = transferRunStatus.TransferredSizeBytes
+		return nil
+	})
+}
+
+func (ts *TransferStateManager) GetDiffHandlingRange(repoKey string, start, end *time.Time) error {
```
eyalbe4, line 186: This comment applies to all getters in this struct, and not only this one.
I suggest we add the value this function retrieves as a returned value of the function. Inn other words, let's change the signature of this function to be - 
```
func (ts *TransferStateManager) GetDiffHandlingRange(repoKey string, start) (end time.Time, err error) {
```

artifactory/commands/transferfiles/status.go
```diff
@@ -0,0 +1,116 @@
...
+	var output strings.Builder
+	defer func() {
+		log.Output(output.String())
+	}()
+
+	startTimestamp, err := state.GetStartTimestamp()
+	if err != nil {
+		return err
+	}
+	if startTimestamp == 0 {
```
eyalbe4, line 27: Question - what happens if thr `transfer-files` command is suddenly stopped? Would the status still show as running? If this is the case, I suggest we remove this "Status" indication for bow, till we develop a way of amking it show the real running state.

artifactory/commands/transferfiles/status.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/jfrog/jfrog-cli-core/v2/utils/coreutils"
+	"github.com/jfrog/jfrog-client-go/utils/log"
+)
+
+const sizeUnits = "KMGTPE"
+
+func ShowStatus() error {
+	var output strings.Builder
+	defer func() {
+		log.Output(output.String())
```
eyalbe4, line 20: What is the reason for printing the output in a defer? Do we really want to print it even in the case of an error in this function?

artifactory/commands/transferfiles/status.go
```diff
@@ -0,0 +1,116 @@
...
+	indentation := strings.Repeat("\t", tabsCount)
+	key += ": "
+	if len(emoji) > 0 {
+		key = emoji + " " + key
+	}
+	output.WriteString(coreutils.PrintBold(key))
+	output.WriteString(indentation + value + "\n")
+}
+
+func sizeToString(sizeInBytes int64) string {
```
eyalbe4, line 109: Since @sarao1310 also needed to handle the format for bytes to string (for the progress bar), it's worth consulting with her whether it's worth consolidating the format logic.

artifactory/commands/transferfiles/transfer.go
```diff
@@ -84,7 +93,24 @@ func (tdc *TransferFilesCommand) SetProxyKey(proxyKey string) {
...
 
+func (tdc *TransferFilesCommand) SetStatus(status bool) {
+	tdc.status = status
+}
+
 func (tdc *TransferFilesCommand) Run() (err error) {
+	if tdc.status {
+		return ShowStatus()
+	}
+	if err := tdc.stateManager.LockTransferStateManager(); err != nil {
```
eyalbe4, line 104: Why we are locking the state manager for the entore duration of the transfer-files process? Is this what this line actually does? I suggest we explain this in a commment.

eyalbe4 (changes requested):

artifactory/commands/transferfiles/status.go
yahavi: Thanks. I added "isProcessRunning" in GetLastLockTimestamp in the latest commit.

artifactory/commands/transferfiles/transfer.go
yahavi: +Added an explanation above LockTransferStateManager method

PR #626, opened 2022-11-17
Transfer config - Optionally check remote repositories URLs

yahavi:
-----

Depends on https://github.com/jfrog/config-import/pull/13
Check remote repositories using the `--prechecks` flag.

The CLI part PR is pending on https://github.com/jfrog/jfrog-cli/pull/1755

### DIFF
--- a/artifactory/commands/transferconfig/transferconfig.go
+++ b/artifactory/commands/transferconfig/transferconfig.go
@@ -2,6 +2,7 @@ package transferconfig
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"net/http"
 	"os"
@@ -39,6 +40,7 @@ type TransferConfigCommand struct {
 	dryRun               bool
 	force                bool
 	verbose              bool
+	preChecks            bool
 	includeReposPatterns []string
 	excludeReposPatterns []string
 	workingDir           string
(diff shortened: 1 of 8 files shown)


artifactory/commands/transferconfig/transferconfig.go
```diff
@@ -319,6 +351,9 @@ func (tcc *TransferConfigCommand) getConfigXml(sourceServiceManager artifactory.
 // Export the config from the source Artifactory to a local directory.
 // Return the path to the export directory, a cleanup function and an error.
 func (tcc *TransferConfigCommand) exportSourceArtifactory(sourceServicesManager artifactory.ArtifactoryServicesManager) (string, func() error, error) {
+	if tcc.preChecks {
```
eyalbe4, line 354: The code in this file performs the `if tcc.preChecks` check multiple times. Since the `--prechecks` option diverts the code into a different flow, which has nothing in common with the code flow without the `--prechecks` option, I'm wondering why can't we run this check once in the beginning, and split the code flow once there.

eyalbe4 (changes requested):

artifactory/commands/transferconfig/transferconfig.go
yahavi: I just tried to do the maximum checks as possible when the `--precheck` flag is provided.
Would you suggest removing all tests that we do regularly in precheck mode?

PR #664, opened 2023-01-04
Use server from Terraform project configuration

RobiNino:
-----

### DIFF
--- a/artifactory/commands/terraform/terraformpublish.go
+++ b/artifactory/commands/terraform/terraformpublish.go
@@ -54,8 +54,14 @@ func (tpc *TerraformPublishCommand) SetArgs(terraformArg []string) *TerraformPub
 	return tpc
 }
 
-func (tpc *TerraformPublishCommand) SetServerDetails(serverDetails *config.ServerDetails) *TerraformPublishCommand {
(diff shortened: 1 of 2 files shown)


artifactory/commands/terraform/terraformpublish.go
```diff
@@ -54,11 +54,17 @@ func (tpc *TerraformPublishCommand) SetArgs(terraformArg []string) *TerraformPub
...
 }
 
-func (tpc *TerraformPublishCommand) SetServerDetails(serverDetails *config.ServerDetails) *TerraformPublishCommand {
+func (tpc *TerraformPublishCommand) setServerDetails(serverDetails *config.ServerDetails) *TerraformPublishCommand {
 	tpc.serverDetails = serverDetails
 	return tpc
 }
 
+func (tpc *TerraformPublishCommand) setRepoConfig(conf *utils.RepositoryConfig) *TerraformPublishCommand {
+	serverDetails, _ := conf.ServerDetails()
```
eyalbe4, line 63: It looks like the signature of the `conf.ServerDetails()` function should be changed to not return an error, beucase the value returned for the error is always nil.

eyalbe4 (changes requested):

artifactory/commands/terraform/terraformpublish.go
RobiNino: This is part of the implementation of the command interface

PR #665, opened 2023-01-10
Convert Repository Filter to Generic Filter

sverdlov93:
-----


artifactory/utils/repositoryutils.go
```diff
@@ -164,18 +165,19 @@ func filterRepositoryNames(repoKeys *[]string, includePatterns, excludePatterns
 	return included, nil
 }
 
-type RepositoryFilter struct {
+type IncludeExcludeFilter struct {
 	IncludePatterns []string
 	ExcludePatterns []string
+	IsRepository    bool
```
eyalbe4, line 171: I think that having the 'IsRepository' isn't the most elegant design.
Maybe we should split the `IncludeExcludeFilter` struct into two structs, with an inheritance relationship between them. Meaning, `RepositoryIncludeExcludeFilter` "extends" `IncludeExcludeFilter` as follows -  
```
type IncludeExcludeFilter struct {
	IncludePatterns []string
	ExcludePatterns []string
}

type RepositoryIncludeExcludeFilter struct {
	IncludeExcludeFilter
}

```

eyalbe4 (changes requested):

PR #675, opened 2023-01-22
Show JWT token subject in 'jf c show'

yahavi:
-----

Add subject to access tokens in `jf c show` command. The subject can be extracted only in JWT access tokens. In reference tokens, for example, the subject is impossible to extract.

image

### DIFF
--- a/common/commands/config.go
+++ b/common/commands/config.go
@@ -538,7 +538,7 @@ func printConfigs(configuration []*config.ServerDetails) {
 		logIfNotEmpty(details.PipelinesUrl, "Pipelines URL:\t\t\t", false, isDefault)
 		logIfNotEmpty(details.User, "User:\t\t\t\t", false, isDefault)
 		logIfNotEmpty(details.Password, "Password:\t\t\t", true, isDefault)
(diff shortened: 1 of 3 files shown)


eyalbe4 (approved):

edgeb1-roche:
Hi,
I'm the individual that originally requested that the token output include some identifying information as to which token was in use.
While it initially appeared that the string provided in the `jf c show` output may in fact be unique to the token in use, and provide some useful information to help identify which token is actually in use, it's not, and it doesn't, and rather, only indicates that you are authenticated to a service, and not, which token is in use, which was the original request.

Using 2 different tokens:

image

The output is identical in both cases:
```
User:                           bruce.edge
Access token:                   *** (Subject: 'jfac@01e54m3ww7cXXXXXXX/users/bruce.edge')
Default:                        true
```

The string provided, redacted above in case there is any external use for it, is identical regardless of which token is in effect at the time.
This was confirmed on a support call with jFrog staff last night for an unrelated problem.
(JFrog Support <> Roche Molecular Systems, Inc. | Case Number: 226528 if that helps)

What would be actually useful, is to show the "Token ID" in the "Access token:..." output.

Can this be changed to show the actual token ID?

PR #680, opened 2023-01-25
Handle API key used as Access Token

sverdlov93:
-----
The API key should be used as a password and not an access token.
The above PR converts the API key to a password when possible and adds a warning.
When it's not possible (no username provided) shows an error.

should fix https://github.com/jfrog/jfrog-cli/issues/1814


common/commands/config.go
```diff
@@ -174,6 +176,9 @@ func (cc *ConfigCommand) config() error {
 		// Some package managers support basic authentication only.
 		// To support them we try to extract the username from the access token
 		if cc.details.AccessToken != "" && cc.details.User == "" {
+			if httpclient.IsApiKey(cc.details.AccessToken) {
+				return errors.New("the provided Access Token is an API key and should be used as a password in username/password authentication")
```
eyalbe4, line 180: Since this error message is duplicated twice in this file, let's move this `errors.New` function into a new function named `newApiKeyUsedAsTokenError()` function and reuse it.

eyalbe4 (changes requested):

PR #705, opened 2023-03-12
Transfer - Reduce snapshot memory consumption

RobiNino:
-----
Memory consumption was reduced by changing the Node struct used by the snapshot manager.
The repository snapshot will not longer save file names, i.e. when resuming from snapshot any directory that wasn't previously completely transferred will be re-explored. This is a breaking change.


artifactory/commands/transferfiles/transfer.go
```diff
@@ -750,3 +750,10 @@ func parseErrorsFromLogFiles(logPaths []string) (allErrors FilesErrors, err erro
 	}
 	return
 }
+
+func assertSupportedTransferDirStructure() error {
+	if err := state.VerifyTransferRunStatusVersion(); err != nil {
+		return err
+	}
+	return assertRepositorySpecificStructure()
```
eyalbe4, line 758: What is the reason we're running the `assertRepositorySpecificStructure ` check? Isn't ti enough running the `VerifyTransferRunStatusVersion`?

eyalbe4 (changes requested):

utils/reposnapshot/node.go
```diff
@@ -11,64 +11,50 @@ import (
...
-	children           map[string]*Node
-	filesNamesAndSizes map[string]int64
-	completed          bool
-	doneExploring      bool
+	mutex    sync.Mutex
+	name     string
+	parent   *Node
+	children []*Node
+	// The files count is used to identify when handling a node is completed. It is only used during runtime, and is not persisted to disk for future runs.
+	filesCount    uint64
```
yahavi, line 19: Can be 32
```suggestion
	filesCount    uint32
```

artifactory/commands/transferfiles/state/utils.go
```diff
@@ -16,6 +16,11 @@ const (
 	secondsInMinute = 60
 	secondsInHour   = 60 * secondsInMinute
 	secondsInDay    = 24 * secondsInHour
+
+	OldTransferDirectoryStructureErrorMsg = "unsupported transfer directory structure found.\n" +
+		"This structure was created by a previous run of the transfer-files command, but is no longer supported by this JFrog CLI version.\n" +
+		"You may either downgrade JFrog CLI to the version that was used before, or remove the transfer directory which is located under the JFrog CLI home directory\n" +
```
yahavi, line 22: "located under the JFrog CLI home directory"
Can we elaborate more about the exact path? I'd prefer that users will be able to work around issues by themselves

utils/reposnapshot/snapshotmanager_test.go
```diff
@@ -192,17 +192,24 @@ func TestAddChildNode(t *testing.T) {
...
-func addAndAssertChild(t *testing.T, childrenMapPool map[string]*Node, root, expectedChild *Node) {
-	assert.NoError(t, root.AddChildNode(expectedChild.name, childrenMapPool))
-	assert.Equal(t, expectedChild, root.children[expectedChild.name])
+func addAndAssertChild(t *testing.T, childrenPool []*Node, root, expectedChild *Node) {
+	assert.NoError(t, root.AddChildNode(expectedChild.name, childrenPool))
+	assert.Equal(t, expectedChild, getChild(root, expectedChild.name))
+}
+
+func getChild(node *Node, childName string) *Node {
+	for i := range node.children {
```
yahavi, line 207: Why not Go style loop?
```suggestion
	for _, child := range node.children {
```

yahavi (changes requested):

PR #712, opened 2023-03-16
Avoid adding the configfile argument to dotnet test command

RobiNino:
-----
Following https://github.com/jfrog/jfrog-cli/issues/1799
`dotnet test` does not support the `--configfile` argument ([open issue](https://github.com/dotnet/sdk/issues/10383)) that was added by default.
Instead, we will now add the `--no-restore` argument to the command to avoid restoring from a different registry that is not Artifactory, and require the user to restore the project beforehand.


artifactory/commands/dotnet/dotnetcommand.go
```diff
@@ -20,6 +21,12 @@ import (
 const (
 	SourceName        = "JFrogCli"
 	configFilePattern = "jfrog.cli.nuget."
+
+	dotnetTestError = `the command failed with an error.
+Note that JFrog CLI does not restore dependencies during a 'dotnet test' command, so if necessary, run a preceding 'dotnet restore'.
```
eyalbe4, line 26: necessary --> needed
("necessary" has a different meaning than "needed", which is like "if you don't have a choice")

artifactory/commands/dotnet/dotnetcommand.go
```diff
@@ -187,11 +197,19 @@ func addSourceToNugetConfig(cmdType dotnet.ToolchainType, configFileName, source
 // Checks if the user provided input such as -configfile flag or -Source flag.
 // If those flags were provided, NuGet will use the provided configs (default config file or the one with -configfile)
 // If neither provided, we are initializing our own config.
-func (dc *DotnetCommand) prepareConfigFile() (cleanup func() error, err error) {
+func (dc *DotnetCommand) prepareConfigFileIfNecessary() (cleanup func() error, err error) {
```
eyalbe4, line 200: prepareConfigFileIfNecessary --> prepareConfigFileIfNeeded

eyalbe4 (changes requested):

PR #779, opened 2023-05-17
Fix curation flaky tests

asafambar:
-----


xray/commands/curation/audit_test.go
```diff
@@ -323,13 +323,15 @@ func Test_treeAnalyzer_getNodesStatusInParallel(t *testing.T) {
 			defer os.Setenv(coreutils.HomeDir, cliHomeDirBefore)
 			currentDir, err := os.Getwd()
 			require.NoError(t, err)
-			require.NoError(t, os.Setenv(coreutils.HomeDir, filepath.Join(currentDir, "../testdata/npm-project/.jfrog")))
+			configurationDir := "../testdata/npm-project/.jfrog"
```
yahavi, line 326:
```suggestion
			configurationDir := filepath.Join("..", "testdata", "npm-project", ".jfrog")
```

yahavi (changes requested):
Looks good!

Please also consider the following, to be consistent with the other tests in this project:
* Remove all underscores from the test names
* Try not to pass 50 lines for a method. This can be achieved by removing the test cases from the tests.
* If possible, avoid negative test loggings. See for example RedirectLogOutputToNil in the JFrog CLI code.

PR #796, opened 2023-05-22
Support creating federated repositories by CLI

sarao1310:
-----


artifactory/commands/repository/repository.go
```diff
@@ -1017,6 +1019,336 @@ func remoteGenericHandler(servicesManager artifactory.ArtifactoryServicesManager
...
+	Alpine:    federatedAlpineHandler,
+	Generic:   federatedGenericHandler,
+}
+
+func federatedMavenHandler(servicesManager artifactory.ArtifactoryServicesManager, jsonConfig []byte, isUpdate bool) error {
+	params := services.NewMavenFederatedRepositoryParams()
+	err := json.Unmarshal(jsonConfig, &params)
+	if errorutils.CheckError(err) != nil {
+		return err
+	}
```
eyalbe4, line 1055: Let's take this opportunity and make the following simplification all accross this file - 
```
	if err := json.Unmarshal(jsonConfig, &params); err != nil {
		return errorutils.CheckError(err)
	}
```

eyalbe4 (changes requested):

artifactory/commands/repository/repository.go
sarao1310: if I simplify like this, err variable will not be defined out of the if condition, and we use it in the below lines, we can change it like this : 

	var err error
	if err = json.Unmarshal(jsonConfig, &params); err != nil {
		return errorutils.CheckError(err)
	}

PR #817, opened 2023-06-01
Improve calculation of Npm dependency tree & Aggregate debug log for max depth reached

EyalDelarea:
-----

This PR is a follow up to https://github.com/jfrog/frogbot/issues/349

There was an infinite loop caused due duplication of indirect dependencies.

On this PR i removed the attempted fix logic and added a unique append to the map of dependencies.

Main:
1. The CLI will show all of the impacted dependencies for each vulnerability.
for example:
Before:
image
After:
image
2. We scan in Npm all of the requested by dependencies. (remove the unclear "depth" const)

**Why there was an infinite loop?**
In a large project (like our webview) we appended to the children a lot of duplicates and it caused the recursion later on to enter an infinite loop. Now with append unique children this won't happen. And we don't have depth restriction.

Update:
This does not effect Frogbot for example as this add a second impact path.


xray/audit/commonutils.go
```diff
@@ -24,17 +24,22 @@ import (
...
-func buildXrayDependencyTree(treeHelper map[string][]string, impactPath []string) *xrayUtils.GraphNode {
+func buildXrayDependencyTree(treeHelper map[string][]string, impactPath []string, exceededDepthCounter *int) *xrayUtils.GraphNode {
 	nodeId := impactPath[len(impactPath)-1]
 	// Initialize the new node
 	xrDependencyTree := &xrayUtils.GraphNode{}
 	xrDependencyTree.Id = nodeId
 	xrDependencyTree.Nodes = []*xrayUtils.GraphNode{}
 	if len(impactPath) >= buildinfo.RequestedByMaxLength {
-		log.Debug("buildXrayDependencyTree exceeded max tree depth")
+		*exceededDepthCounter += 1
```
eyalbe4, line 42: *exceededDepthCounter += 1
-->
*exceededDepthCounter++

eyalbe4 (changes requested):

EyalDelarea:
@eyalbe4 I've added before and after screenshots.
Just wanted to make sure we still want this output so i'll merge this.
i've tested it, and it does not effect Frogbot, trying to think if anything else would change from this new output.
Basically it's just another impact path so i shouldn't.

eyalbe4 (approved):

PR #823, opened 2023-06-08
Transfer - Disable repo snapshot for delayed files

RobiNino:
-----
Before this change, delayed files were tracked in the transfer repository snapshot (by incrementing the files count). This could block potential tree collapsing in the snapshot tree, since the directories containing those files would only be completed in later stage. Not collapsing the tree could in turn cause an increase in memory.

### DIFF
--- a/artifactory/commands/transferfiles/fulltransfer.go
+++ b/artifactory/commands/transferfiles/fulltransfer.go
@@ -100,7 +100,14 @@ func (m *fullTransferPhase) run() error {
 		_, err := pcWrapper.chunkBuilderProducerConsumer.AddTaskWithError(folderHandler(folderParams{relativePath: "."}), pcWrapper.errorsQueue.AddError)
 		return err
 	}
-	delayAction := consumeDelayFilesIfNoErrors
+	delayAction := func(phase phaseBase, addedDelayFiles []string) error {
+		// Disable repo transfer snapshot as it is not used for delayed files.
+		if err := m.stateManager.SaveStateAndSnapshots(); err != nil {
+			return err
+		}
+		m.stateManager.DisableRepoTransferSnapshot()
+		return consumeDelayFilesIfNoErrors(phase, addedDelayFiles)
+	}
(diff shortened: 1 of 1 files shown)


artifactory/commands/transferfiles/fulltransfer.go
```diff
@@ -100,7 +100,14 @@ func (m *fullTransferPhase) run() error {
 		_, err := pcWrapper.chunkBuilderProducerConsumer.AddTaskWithError(folderHandler(folderParams{relativePath: "."}), pcWrapper.errorsQueue.AddError)
 		return err
 	}
-	delayAction := consumeDelayFilesIfNoErrors
+	delayAction := func(phase phaseBase, addedDelayFiles []string) error {
+		// Disable repo transfer snapshot as it is not used for delayed files.
+		if err := m.stateManager.SaveStateAndSnapshots(); err != nil {
+			return err
+		}
+		m.stateManager.DisableRepoTransferSnapshot()
```
yahavi, line 108: I'd put this code above the invocation of the delay action, to emphasize that the "DisableRepoTransferSnapshot" is called after the transfer:
```go
if returnedError == nil && delayAction != nil {
	// Disable repo transfer snapshot as it is not used for delayed files.
	if err := ftm.stateManager.SaveStateAndSnapshots(); err != nil {
		return err
	}
	ftm.stateManager.DisableRepoTransferSnapshot()

	var addedDelayFiles []string
	// If the transfer generated new delay files provide them
	if delayedArtifactsMng.delayedWriter != nil {
		addedDelayFiles = delayedArtifactsMng.delayedWriter.contentFiles
	}
	returnedError = delayAction(ftm.phaseBase, addedDelayFiles)
}
```
And all of this can be exported to a new method "handleDelayedArtifacts".

yahavi (changes requested):

artifactory/commands/transferfiles/fulltransfer.go
RobiNino: This code is only relevant to the full transfer phase, so I don't think we should add it to the general manager functionality.

PR #826, opened 2023-06-13
Lower transfer retry time

RobiNino:
-----
Previous total retry time was mistakenly raised too high. This PR sets it to be 1 hour.

### DIFF
--- a/artifactory/commands/transferfiles/transfer.go
+++ b/artifactory/commands/transferfiles/transfer.go
@@ -29,8 +29,8 @@ const (
 	uploadChunkSize = 16
 	// Size of the channel where the transfer's go routines write the transfer errors
 	fileWritersChannelSize       = 500000
-	retries                      = 10000
-	retriesWaitMilliSecs         = 5000
+	retries                      = 1800
+	retriesWaitMilliSecs         = 2000
 	dataTransferPluginMinVersion = "1.7.0"
 )
 


artifactory/commands/transferfiles/transfer.go
```diff
@@ -29,8 +29,8 @@ const (
 	uploadChunkSize = 16
 	// Size of the channel where the transfer's go routines write the transfer errors
 	fileWritersChannelSize       = 500000
-	retries                      = 10000
-	retriesWaitMilliSecs         = 5000
+	retries                      = 1800
+	retriesWaitMilliSecs         = 2000
```
yahavi, line 33: (2 seconds + 1 second for the REST) X 1800 retries = 1.5 hours. Are we OK with that?

yahavi, line 33: Retries happen when a system is experiencing significant stress. Using too low backoff time would make it worse.
I suggest keeping the 5 seconds and adjust the number of retries accordingly.

yahavi (changes requested):

artifactory/commands/transferfiles/transfer.go
```diff
@@ -29,7 +29,7 @@ const (
 	uploadChunkSize = 16
 	// Size of the channel where the transfer's go routines write the transfer errors
 	fileWritersChannelSize       = 500000
-	retries                      = 10000
+	retries                      = 720
```
yahavi, line 32: This is more than an hour. 72 seconds.

PR #834, opened 2023-06-21
Support environments in repository creation

RobiNino:
-----
Adds the `environments` option to the `jf rt rpt` command

### DIFF
--- a/artifactory/commands/repository/template.go
+++ b/artifactory/commands/repository/template.go
@@ -19,11 +19,8 @@ type RepoTemplateCommand struct {
 }
 
 const (
-	PathErrorSuffixMsg = " please enter a path, in which the new template file will be created"
-
 	// Strings for prompt questions
-	SelectConfigKeyMsg   = "Select the next configuration key" + utils.PressTabMsg
-	InsertValuePromptMsg = "Insert the value for "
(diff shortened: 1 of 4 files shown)


artifactory/commands/repository/template.go
```diff
@@ -798,6 +797,11 @@ var questionMap = map[string]utils.QuestionInfo{
 		Writer:    utils.WriteStringAnswer,
 		Callback:  projectKeyCallback,
 	},
+	Environments: {
+		PromptPrefix: "Insert the name of the environment to assign to >",
```
yahavi, line 801: Since "environments" is plural, I guessed there is a way to provide multiple environments.
This was my first attempt:
image

How can one configure multiple environments?

yahavi (changes requested):

artifactory/commands/repository/template.go
RobiNino: You can't, which is why the prompt only offers to insert one environment.
Perhaps the repository configuration was built this way to support this in the future, but currently the REST API yields an error if you provide more than one environment (the UI behaves the same).

yahavi: Can it be changed to "environment"?

PR #836, opened 2023-06-25
Exclude JAS Scan Flag

orto17:
-----
 
This PR contains the following:
* A new flag for the audit command hat will allow the user to force a skip on a certain Jas scan.
* Each scan that will appear in the flag will be skipped, and analyzer manager will not be executed for this scan.
* The user will be able to include more than one scanner to skip, for example: `-exclude-scan=contextual_analysis;secrets`

### DIFF
--- a/xray/audit/jas/jasmanager.go
+++ b/xray/audit/jas/jasmanager.go
@@ -10,6 +10,7 @@ import (
 	"github.com/jfrog/jfrog-client-go/xray/services"
 	xrayUtils "github.com/jfrog/jfrog-client-go/xray/services/utils"
 	"os"
+	"strings"
 )
 
 const serverDetailsErrorMessage = "cant get xray server details"
@@ -17,10 +18,11 @@ const serverDetailsErrorMessage = "cant get xray server details"
 var (
 	analyzerManagerExecuter utils.AnalyzerManagerInterface = &utils.AnalyzerManager{}
 	skippedDirs                                            = []string{"**/*test*/**", "**/*venv*/**", "**/*node_modules*/**", "**/*target*/**"}
+	scannersToExclude                                      = []string{}
 )
 
 func GetExtendedScanResults(xrayResults []services.ScanResponse, dependencyTrees []*xrayUtils.GraphNode,
-	serverDetails *config.ServerDetails) (*utils.ExtendedScanResults, error) {
+	serverDetails *config.ServerDetails, excludeScan string) (*utils.ExtendedScanResults, error) {
 	if serverDetails == nil {
 		return nil, errors.New(serverDetailsErrorMessage)
 	}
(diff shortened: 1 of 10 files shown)


xray/audit/jas/jasmanager.go
```diff
@@ -10,17 +10,19 @@ import (
...
 	"os"
+	"strings"
 )
 
 const serverDetailsErrorMessage = "cant get xray server details"
 
 var (
 	analyzerManagerExecuter utils.AnalyzerManagerInterface = &utils.AnalyzerManager{}
 	skippedDirs                                            = []string{"**/*test*/**", "**/*venv*/**", "**/*node_modules*/**", "**/*target*/**"}
+	scannersToExclude                                      = []string{}
```
yahavi, line 21: This comment is not related to this PR, but is more general - 
Please prefer not to use global variables. Instead, the recommended way is to use a new JasManager struct.

One design pattern is to pass a pointer of the struct when needed to each one of the scanners. 
Another design pattern is to use inheritance/decoration, whereby every scanner contains the JasManager struct.

xray/audit/jas/jasmanager.go
```diff
@@ -39,6 +41,7 @@ func GetExtendedScanResults(xrayResults []services.ScanResponse, dependencyTrees
 	if err = utils.CreateAnalyzerManagerLogDir(); err != nil {
 		return nil, err
 	}
+	scannersToExclude = strings.Split(excludeScan, ";")
```
yahavi, line 44: Let's also trim strings to allow for example: `--exclude-scan=contextual_analysis; secrets`

xray/audit/jas/secretsscanner.go
```diff
@@ -49,6 +50,9 @@ func getSecretsScanResults(serverDetails *config.ServerDetails, analyzerManager
 			err = errors.Join(err, cleanupFunc())
 		}
 	}()
+	if utils.ExcludeScan(scannersToExclude, secretsFeatureName) {
+		return nil, false, nil
```
yahavi, line 54: Let's add a debug level log that the secret scan is skipped

xray/utils/analyzermanager.go
```diff
@@ -243,3 +243,12 @@ func GetResultSeverity(result *sarif.Result) string {
...
 	return SeverityDefaultValue
 }
+
+func ExcludeScan(scansToBeExcluded []string, scan string) bool {
+	for _, s := range scansToBeExcluded {
+		if s == scan {
+			return true
+		}
+	}
+	return false
```
yahavi, line 253: Can be inlined:
```suggestion
	return slices.Contains(scansToBeExcluded, scan)
```

yahavi (changes requested):
Looks good @orto17!
Please consider my inline comments.

PR #867, opened 2023-07-17
New `--module` option added to the `build-add-dependencies` command

sarao1310:
-----


artifactory/commands/buildinfo/adddependencies.go
```diff
@@ -286,6 +286,9 @@ func (badc *BuildAddDependenciesCommand) savePartialBuildInfo(dependencies []bui
 	populateFunc := func(partial *buildinfo.Partial) {
 		partial.ModuleType = buildinfo.Generic
 		partial.Dependencies = dependencies
+		if badc.buildConfiguration.GetModule() != "" {
```
eyalbe4, line 289: Is this if statement actually needed? Why can't we set the value without having this check first?

eyalbe4 (changes requested):

artifactory/commands/buildinfo/adddependencies.go
sarao1310: Yes, we do. we have been following a specific approach until now, which involves generating a partial with a module ID equal to the build name when the user does not make a module ID selection.
without this if statement, is the user does not select a module ID, the module ID will be an empty string.

PR #880, opened 2023-07-28
Transfer files - Fix `repo snapshot manager` error

eyalbe4:
This fix resolves an issue with the `jf rt transfer-files` command. 
The symptom of this issue is the following error message that occasionally could be seen in the log:
```
repo snapshot manager - could not reach the representing node for path...
```
The key for fixing the issue, is the change made to the `findMatchingNode` function. The issue was that the `findMatchingNode` function could not always return the node. This is because the node may not be in the snapshot cache. It may yet to be discovered, or maybe it was discovered, but the snapshot cache didn't get a chance to get persisted yet, because of a shutdown of the command.

With this fix, the `findMatchingNode` function always fetches the node, It returns it if it's cached, and creates it and adds it to the cache if it isn't cached. As a result, the `AddChildNode` function became redundant and was left only to be used by one of tests,


talarian1:
Duplicate https://github.com/jfrog/jfrog-cli-core/pull/881

PR #907, opened 2023-08-22
Improve Audit and SCA scan code and fix error aggregation 

sverdlov93:
-----

### DIFF
--- a/xray/commands/audit/generic/auditmanager.go
+++ b/xray/commands/audit/generic/auditmanager.go
@@ -4,9 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"github.com/jfrog/build-info-go/utils/pythonutils"
-	"github.com/jfrog/gofrog/version"
 	rtutils "github.com/jfrog/jfrog-cli-core/v2/artifactory/utils"
-	"github.com/jfrog/jfrog-cli-core/v2/utils/config"
 	"github.com/jfrog/jfrog-cli-core/v2/utils/coreutils"
 	"github.com/jfrog/jfrog-cli-core/v2/xray/audit"
 	_go "github.com/jfrog/jfrog-cli-core/v2/xray/audit/go"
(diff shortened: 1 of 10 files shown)


xray/commands/audit/generic/auditmanager.go
```diff
@@ -126,20 +124,27 @@ func RunAudit(auditParams *Params) (results *Results, err error) {
...
+
 	auditParams.SetXrayVersion(xrayVersion)
 
 	errGroup := new(errgroup.Group)
 	if isEntitled {
 		// Download (if needed) the analyzer manager in a background routine.
 		errGroup.Go(rtutils.DownloadAnalyzerManagerIfNeeded)
 	}
-
 	// The audit scan doesn't require the analyzer manager, so it can run separately from the analyzer manager download routine.
```
omerzi, line 146:
```suggestion
	// The sca scan doesn't require the analyzer manager, so it can run separately from the analyzer manager download routine.
```

xray/commands/audit/generic/auditmanager.go
```diff
@@ -154,88 +159,69 @@ func RunAudit(auditParams *Params) (results *Results, err error) {
...
+	results = NewAuditResults()
+	rootDir, err := os.Getwd()
+	if errorutils.CheckError(err) != nil {
+		return results.SetAuditError(err)
 	}
-	defer func() {
-		err = errors.Join(err, os.Chdir(projectDir))
-	}()
-	results := NewAuditResults()
 	for _, wd := range params.workingDirs {
```
omerzi, line 177: In case we don't supply a specific `workingDirs`, does the root dir is a part of the `workingDirs` slice?

xray/commands/audit/generic/generic.go
```diff
@@ -75,15 +75,19 @@ func (auditCmd *GenericAuditCommand) CreateXrayGraphScanParams() *services.XrayG
 }
 
 func (auditCmd *GenericAuditCommand) Run() (err error) {
+	workingDirs, err := xrutils.GetFullPathsWorkingDirs(auditCmd.workingDirs)
```
omerzi, line 78: I believe this logic should reside within the inner audit logic since it is applicable only to the CLI context. While it is true that we also address this case in Frogbot, we should not be surprised by this behavior for future features that utilize Audit. (In cases where the working directory is empty, the scan won't automatically run on the root directory.)

PR #908, opened 2023-08-23
Add usage reporter

attiasas:
-----

### DIFF
--- a/utils/usage/usage.go
+++ b/utils/usage/usage.go
@@ -1,0 +1,228 @@
+package usage
+
+import (
+	"golang.org/x/sync/errgroup"
+
+	"github.com/jfrog/jfrog-cli-core/v2/artifactory/utils"
+	"github.com/jfrog/jfrog-cli-core/v2/utils/config"
+	"github.com/jfrog/jfrog-cli-core/v2/utils/coreutils"
+	xray "github.com/jfrog/jfrog-cli-core/v2/xray/commands/utils"
+	"github.com/jfrog/jfrog-client-go/artifactory/usage"
+	clientutils "github.com/jfrog/jfrog-client-go/utils"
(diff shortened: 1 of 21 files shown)


utils/usage/usage.go
```diff
@@ -0,0 +1,225 @@
...
+		ProductId:         productId,
+		serverDetails:     serverDetails,
+		reportWaitGroup:   new(errgroup.Group),
+		sendToEcosystem:   true,
+		sendToXray:        true,
+		sendToArtifactory: true,
+	}
+}
+
+func ShouldReportUsage() (reportUsage bool) {
```
eyalbe4, line 55: Let's use this function also in -
```
func reportUsage(command Command, channel chan<- bool) {
	defer signalReportUsageFinished(channel)
	reportUsage, err := clientutils.GetBoolEnvValue(coreutils.ReportUsage, true)
	if err != nil {
		log.Debug(usageReporter.ReportUsagePrefix, err.Error())
		return
	}
```
instead of calling `clientutils.GetBoolEnvValue` directly.

Also, why don't we remove the above `reportUsage` function and use the new UsageReporter instead?

utils/usage/usage.go
```diff
@@ -0,0 +1,225 @@
...
+			return ur.reportToArtifactory(features...)
+		})
+	}
+}
+
+func (ur *UsageReporter) WaitForResponses() (err error) {
+	return ur.reportWaitGroup.Wait()
+}
+
+func (ur *UsageReporter) reportToEcosystem(features ...ReportFeature) (err error) {
```
eyalbe4, line 107: Let's add the `ReportUsagePrefix` to all the errors generated by this function, as well as the `reportToXray` and `reportToArtifactory` functions.

eyalbe4 (changes requested):

PR #938, opened 2023-09-07
Support Platform Access Token

RobiNino:
-----
Depends on https://github.com/jfrog/jfrog-client-go/pull/822


general/token/accesstokencreate.go
```diff
@@ -0,0 +1,150 @@
+package generic
```
yahavi, line 1:
```suggestion
package login
```

yahavi, line 1: Let's rename this file to "createaccesstoken"

general/token/accesstokencreate.go
```diff
@@ -0,0 +1,150 @@
...
+	content, err := json.Marshal(*atc.response)
+	return content, errorutils.CheckError(err)
+}
+
+func (atc *AccessTokenCreateCommand) ServerDetails() (*config.ServerDetails, error) {
+	return atc.serverDetails, nil
+}
+
+func (atc *AccessTokenCreateCommand) CommandName() string {
+	return "access_token_create"
```
yahavi, line 106:
```suggestion
	return "create_access_token"
```

yahavi (changes requested):

general/token/accesstokencreate.go
RobiNino: Not part of the login mechanism. Changed to `token`

RobiNino: I think we should follow the convention where the file is named after the command it adds.

PR #946, opened 2023-09-12
Enhance the `transfer-setting` command to allow modifying the log level

eyalbe4:
Emhance the `transfer-setting` command to allow modifying the log level while the `transfer-files` command is running.

Depends on https://github.com/jfrog/jfrog-client-go/pull/828


artifactory/commands/transfer/settings.go
```diff
@@ -25,28 +28,60 @@ func (tst *TransferSettingsCommand) Run() error {
...
+	if currLogLevel == "" {
+		currLogLevel = os.Getenv(coreutils.LogLevel)
+	}
+	if currLogLevel == "" {
+		currLogLevel = "INFO"
+	}
+	return currLogLevel
+}
+
+func (tst *TransferSettingsCommand) validateLogLevelValue(loglevel string) error {
```
yahavi, line 78:
```suggestion
func (tst *TransferSettingsCommand) validateLogLevelValue(logLevel string) error {
```

artifactory/utils/transfersettings.go
```diff
@@ -22,7 +22,8 @@ const (
 )
 
 type TransferSettings struct {
-	ThreadsNumber int `json:"threadsNumber,omitempty"`
+	ThreadsNumber int    `json:"threadsNumber,omitempty"`
+	LogLevel      string `json:"logLevel,omitempty"`
```
yahavi, line 26: I think it will be safer to work with the same log-level type, using our methods for handling the log levels.
```suggestion
	LogLevel      log.LevelType `json:"logLevel,omitempty"`
```

yahavi (changes requested):

artifactory/commands/transfer/settings.go
```diff
@@ -25,28 +28,60 @@ func (tst *TransferSettingsCommand) Run() error {
...
+	if err = utils.SaveTransferSettings(conf); err != nil {
 		return err
 	}
 	log.Output("The settings were saved successfully. It might take a few moments for the new settings to take effect.")
 	log.Output(fmt.Sprintf("Note - For Build Info repositories, the number of worker threads will be limited to %d.", utils.MaxBuildInfoThreads))
 	return nil
 }
 
+func (tst *TransferSettingsCommand) getCurrLogLevel(settings utils.TransferSettings) string {
+	currLogLevel := settings.LogLevel
```
RobiNino, line 68: Missing a check if `settings != nil` here.

artifactory/commands/transfer/settings.go
```diff
@@ -25,28 +28,60 @@ func (tst *TransferSettingsCommand) Run() error {
...
+		currLogLevel = os.Getenv(coreutils.LogLevel)
+	}
+	if currLogLevel == "" {
+		currLogLevel = "INFO"
+	}
+	return currLogLevel
+}
+
+func (tst *TransferSettingsCommand) validateLogLevelValue(loglevel string) error {
+	if loglevel != "DEBUG" && loglevel != "INFO" && loglevel != "WARN" && loglevel != "ERROR" {
```
RobiNino, line 79: I suggest a cleaner approach:
```
	switch loglevel {
	case "DEBUG", "INFO", "WARN", "ERROR":
		return nil
	default:
		return errorutils.CheckErrorf("the log level value is invalid")
	}

**Review by RobiNino — CHANGES_REQUESTED** [review]

File: artifactory/commands/transferfiles/utils.go [hunk 5]
```diff
@@ -338,17 +339,37 @@ func periodicallyUpdateThreadsAndStopStatus(pcWrapper *producerConsumerWrapper,
 			log.Debug("Stopping the polling on the settings and stop files for the current phase.")
 			return
 		}
-		if err := updateThreads(pcWrapper, buildInfoRepo); err != nil {
+
+		settings, err := utils.LoadTransferSettings()
+		if err != nil || settings == nil {
```
RobiNino, line 344: We should not return if settings is `nil`, we should `continue`. Settings file does not exist on a clean run, until it is created with the settings command.

So far we also didn't return here if an error occurred. Our approach was to log it and not stop the transfer due to an error in the settings. If we decide to change approach, we should change it in all relevant places.

In addition, only log the error if it is not `nil`, this currently logs the error also when both the settings and error is `nil`.

RobiNino (changes requested):

PR #961, opened 2023-09-13
Transfer - Fix delayed artifacts mechanism

yahavi:
-----

Tests PR: https://github.com/jfrog/jfrog-cli/pull/2204

Enhance and optimize the following elements:
   - Delaying pom.xml files, not just `*.pom`.
   - Eliminating the need for recursion in delaying artifacts. Instead, transfer them in a sorted manner. This approach is significantly more efficient, especially considering that in Docker and Conan repositories, we triggered 2-3 invocations of doTransferWithProducerConsumer.

### DIFF
--- a/artifactory/commands/transferfiles/delayedartifactshandler.go
+++ b/artifactory/commands/transferfiles/delayedartifactshandler.go
@@ -97,7 +97,7 @@ func consumeAllDelayFiles(base phaseBase) error {
 	delayFunctions := getDelayUploadComparisonFunctions(base.repoSummary.PackageType)
 	if len(filesToConsume) > 0 && len(delayFunctions) > 0 {
 		log.Info("Starting to handle delayed artifacts uploads...")
-		if err = handleDelayedArtifactsFiles(filesToConsume, base, delayFunctions[1:]); err == nil {
+		if err = handleDelayedArtifactsFiles(filesToConsume, base, delayFunctions); err == nil {
 			log.Info("Done handling delayed artifacts uploads.")
 		}
 	}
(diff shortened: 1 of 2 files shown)


artifactory/commands/transferfiles/delayedartifactshandler.go
```diff
@@ -168,7 +160,8 @@ func consumeDelayedArtifactsFiles(pcWrapper *producerConsumerWrapper, filesToCon
 			return err
 		}
 
-		shouldStop, err := uploadByChunks(delayedArtifactsFile.DelayedArtifacts, uploadChunkChan, base, delayHelper, errorsChannelMng, pcWrapper)
+		sortedDelayedArtifacts := sortDelayedArtifacts(delayedArtifactsFile, delayHelper)
```
RobiNino, line 163: I believe this is only sorting the artifacts within a specific delayedArtifactsFile. Those files are handled in parallel, meaning that even if a single file is being handled in the correct order, artifacts from different files can be uploaded in the wrong order.

artifactory/commands/transferfiles/delayedartifactshandler.go
```diff
@@ -246,8 +255,10 @@ type delayUploadHelper struct {
 }
 
 // Decide whether to delay the deployment of a file by running over the shouldDelayUpload array.
-// When there are multiple levels of requirements in the deployment order, the first comparison function in the array can be removed each time in order to no longer delay by that rule.
 func (delayHelper delayUploadHelper) delayUploadIfNecessary(phase phaseBase, file api.FileRepresentation) (delayed, stopped bool) {
+	if phase.phaseId != api.Phase1 {
```
RobiNino, line 259: Shouldn't files be delayed in other phases? For example:
Phase 2 - Found new docker image that contains a `manifest.json`.
Phase 3 - Handling failures, `manifest.json` can be one of them.

RobiNino (changes requested):

artifactory/commands/transferfiles/delayedartifactshandler.go
yahavi: Good catch! Thanks

artifactory/commands/transferfiles/delayedartifactshandler.go
yahavi: You are right. Thanks

artifactory/commands/transferfiles/delayedartifactshandler.go
```diff
@@ -149,32 +149,38 @@ func handleDelayedArtifactsFiles(filesToConsume []string, base phaseBase, delayU
...
-		shouldStop, err := uploadByChunks(delayedArtifactsFile.DelayedArtifacts, uploadChunkChan, base, delayHelper, errorsChannelMng, pcWrapper)
+		allDelayedArtifacts = append(allDelayedArtifacts, delayedArtifactsFile.DelayedArtifacts...)
+	}
+
+	// Upload delayed artifacts, if any exist
+	if len(allDelayedArtifacts) > 0 {
+		log.Info(fmt.Sprintf("Uploading %d delayed artifacts...", len(allDelayedArtifacts)))
+		allDelayedArtifacts = sortDelayedArtifacts(&allDelayedArtifacts, delayHelper)
+		// Use an empty delayUploadHelper to prevent further delays of these files.
+		shouldStop, err := uploadByChunks(allDelayedArtifacts, uploadChunkChan, base, delayUploadHelper{}, errorsChannelMng, pcWrapper)
```
attiasas, line 174: If a delayed artifact fails here, for example, 'conanfile.py', then the following file at the ordered delayed list is for example 'conaninfo.txt' and no issue at upload, won't it be a problem later when we try to upload It again?

PR #976, opened 2023-09-27
Build flat dependency trees while auditing Gradle projects

asafgabai:
-----
A new, more memory-efficient structure for the dependency trees of Gradle projects.


xray/commands/audit/sca/java/javautils.go
```diff
@@ -173,3 +176,73 @@ func (dm *dependencyMultimap) putChild(parent string, child *buildinfo.Dependenc
 func (dm *dependencyMultimap) getChildren(parent string) map[string]*buildinfo.Dependency {
 	return dm.multimap[parent]
 }
+
+// The structure of a dependency tree of a module in a Gradle/Maven project, as created by the gradle-dep-tree and maven-dep-tree plugins.
+type moduleDepTree struct {
+	Root  string                 `json:"root"`
+	Nodes map[string]depTreeNode `json:"nodes"`
```
yahavi, line 183: Let's add "omitempty" to ensure we don't fail projects without children
```suggestion
	Nodes map[string]depTreeNode `json:"nodes,omitempty"`
```

xray/commands/audit/sca/java/javautils.go
```diff
@@ -173,3 +176,73 @@ func (dm *dependencyMultimap) putChild(parent string, child *buildinfo.Dependenc
...
+}
+
+type depTreeNode struct {
+	Children []string `json:"children"`
```
yahavi, line 187:
```suggestion
	Children []string `json:"children,omitempty"`
```

xray/commands/audit/sca/java/javautils.go
```diff
@@ -173,3 +176,73 @@ func (dm *dependencyMultimap) putChild(parent string, child *buildinfo.Dependenc
...
+	for _, childId := range moduleTree.Nodes[currNodeId].Children {
+		childGav := GavPackageTypeIdentifier + childId
+		childNode := &xrayUtils.GraphNode{
+			Id:     childGav,
+			Nodes:  []*xrayUtils.GraphNode{},
+			Parent: currNode,
+		}
+		if currNode.NodeHasLoop() {
+			return
+		}
```
yahavi, line 220: Let's put it in the beginning of the loop

yahavi (changes requested):
Top-level @asafgabai!
Please consider my inline comments.

xray/commands/audit/sca/java/javautils.go
asafgabai: It's not necessary to use `omitempty` in this context, as `omitempty` is typically used when encoding Go objects to JSON to omit empty fields. In our case, we only parse the plugins' JSON results into Go objects, where missing fields are parsed as having their empty value.

xray/commands/audit/sca/java/javautils.go
asafgabai: I moved it outside of the loop.

PR #982, opened 2023-09-29
fix(build-add-git): severe performance issue

gerrnot:
On our artifactory instance, the performance of the build-add-git command increased from ~4m to <1s, even for repositories with 20k commits

the functionality `getLatestVcsRevision` from (possible) previously uploaded build infos was removed for the following reasons:
- the aql query was broken anyway (result was always empty)
- the broken aql query lead to a call to a GET call to /api/build without further args, which seems to retrieve all builds and caused severe performance issues
- other than for performance reasons, I see no significant reason to query the vcs revision history of past builds -> And since almost any api call will always be slower than just scanning the whole git history of a repo, I don't see the reason for this functionality to be there

-----


gerrnot:

gerrnot:
this seems having been fixed independently in meantime -> successfully tested it with jf cli version `2.52.9`

PR #1010, opened 2023-10-26
Transfer - Interruption may causes data loss

yahavi:
-----

Halting the "transfer-files" process while delayed artifacts are being transferred might result in the early removal of the delayed artifacts file before all the chunks are properly uploaded. As a result, these files won't be uploaded unless you employ the `--ignore-state` flag. A similar problem can occur during the handling of error files.

This image serves as an illustration of the problem:
image

### DIFF
--- a/artifactory/commands/transferfiles/delayedartifactshandler.go
+++ b/artifactory/commands/transferfiles/delayedartifactshandler.go
@@ -89,19 +89,21 @@ type DelayedArtifactsFile struct {
 }
 
 // Collect all the delayed artifact files that were created up to this point for the repository and transfer their artifacts using handleDelayedArtifactsFiles
-func consumeAllDelayFiles(base phaseBase) error {
+func consumeAllDelayFiles(base phaseBase) (err error) {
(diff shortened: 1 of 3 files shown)


artifactory/commands/transferfiles/delayedartifactshandler.go
```diff
@@ -95,13 +95,17 @@ func consumeAllDelayFiles(base phaseBase) error {
...
+
+	log.Info("Starting to handle delayed artifacts uploads...")
+	if err = handleDelayedArtifactsFiles(filesToConsume, base, delayFunctions[1:]); err != nil {
+		return err
 	}
-	return err
+
+	log.Info("Done handling delayed artifacts uploads.")
+	return deleteAllFiles(filesToConsume)
 }
```
eyalbe4, line 109: Let's add a comment explaining why the first element is skipped in `delayFunctions[1:]` in the following line in inside this function.
```
if err = handleDelayedArtifactsFiles(filesToConsume, base, delayFunctions[1:]); err != nil {
```

PR #1031, opened 2023-11-14
Support for Distribution path mapping with AQL

YardenEdery:
-----


YardenEdery:

YardenEdery:
Not sure about adding tests here, there are none related to distribution

common/spec/specfiles.go
```diff
@@ -194,6 +197,9 @@ func ValidateSpec(files []File, isTargetMandatory, isSearchBasedSpec bool) error
 		isBypassArchiveInspection, _ := file.IsBypassArchiveInspection(false)
 		isTransitive, _ := file.IsTransitive(false)
 
+		if isPathMapping && !isAql {
+			return errorutils.CheckErrorf("mappings supported only with Aql")
```
eyalbe4, line 201: mappings supported only with Aql
-->
path mapping is supported only with Aql

common/spec/specfiles.go
```diff
@@ -206,6 +212,12 @@ func ValidateSpec(files []File, isTargetMandatory, isSearchBasedSpec bool) error
 		if isSearchBasedSpec && !isAql && !isPattern && !isBuild && !isBundle {
 			return errorutils.CheckErrorf("spec must include either aql, pattern, build or bundle")
 		}
+		if isPathMapping && isTarget {
+			return fileSpecValidationError("mappings", "target")
+		}
+		if isPathMapping && isPattern {
+			return fileSpecValidationError("mappings", "pattern")
+		}
```
eyalbe4, line 220: Let's place the above line right after the following code for better readability - 
```
if isPathMapping && !isAql {
	return errorutils.CheckErrorf("mappings supported only with Aql")
}
```

eyalbe4 (changes requested):
Let's add a complete description for this PR.

common/spec/specfiles.go
YardenEdery: Are you sure because "mappings" is the interface to the user, they will use "mappings" in their spec.json and I saw in the code you treat it as PathMappings so I kept the same conventions, and thought that the error should show the user what he needs to fix.

eyalbe4: OK. NP.

common/spec/specfiles.go
```diff
@@ -41,6 +41,7 @@ func CreateSpecFromFile(specFilePath string, specVars map[string]string) (spec *
 
 type File struct {
 	Aql                     utils.Aql
+	PathMapping             utils.PathMapping `json:"mappings,omitempty"`
```
yahavi, line 44: Can we leave it "pathMapping"? Or use "mapping"?
```suggestion
	PathMapping             utils.PathMapping
```

yahavi (changes requested):

common/spec/specfiles.go
YardenEdery: I tried to keep the convention, I saw in lifecycle directory the use of pattern/target (which later on transform to "mappings") as "PathMappings", thus tried to keep it the same. I can change it to `Mappings` in the `File struct` but eventually its similar with lifecycles "PathMapping" which I preferred not to touch / delete / modify.
WDYT?
youll see what im talking about by searching in the code (jfrog-client-go):
```
 type Modifications struct {
	PathMappings []utils.PathMapping `json:"mappings"`
}

type PathMapping struct {
	Pattern string
	Target  string
}
```

I do think that changing to "mappings" will be better because thats how distribution server expects it eventually

yahavi (dismissed):
LGTM

PR #1040, opened 2023-11-20
.Net/NuGet- Enhancements to Artifactory Integration and auto-installation for .NET/NuGet projects

eranturgeman:
-----

Summery of changes:

Implemented the ability to configure an Artifactory server as a resolution server while constructing the dotnet/nuget dependency tree.

Introduced auto-install functionality for dotnet/nuget projects within the process of building dependency trees. Users can also supply customized installation commands. Note that not all entry points support the provision of these parameters.

Resolved a bug related to enabling an Artifactory server to be configured as a resolution server for projects that were configured using 'jf dotnet-config'

Added tests and test cases for the newly introduced functionalities.


xray/commands/audit/sca/nuget/nuget.go
```diff
@@ -62,20 +62,23 @@ func BuildDependencyTree(params utils.AuditParams) (dependencyTree []*xrayUtils.
 	return
 }
 
+// Verifies whether the execution of an 'install' command is necessary, either because the project isn't installed or because the user has specified an 'install' command
 func isInstallRequired(params utils.AuditParams, sol solution.Solution) (installRequired bool) {
```
omerzi, line 66:
```suggestion
func isInstallRequired(params utils.AuditParams, sol solution.Solution) bool {
	// If the user has specified an 'install' command, we proceed with executing the 'restore' command even if the project is already installed
	// Additionally, if dependency sources were not identified during the construction of the Solution struct, the project will necessitate an 'install'
	return len(params.InstallCommandArgs()) > 0 || !sol.DependenciesSourcesAndProjectsPathExist()
}

File: xray/commands/audit/sca/nuget/nuget.go [hunk 2]
```diff
@@ -62,20 +62,23 @@ func BuildDependencyTree(params utils.AuditParams) (dependencyTree []*xrayUtils.
...
 	return
 }
 
+// Generates a temporary duplicate of the project to execute the 'install' command without impacting the original directory and establishing the JFrog configuration file for Artifactory resolution
+// Additionally, re-loads the project's Solution so the dependencies sources will be identified
 func runDotnetRestoreAndLoadSolution(params utils.AuditParams, originalWd string) (sol solution.Solution, err error) {
 	// Creating a temporary copy of the project in order to run 'install' command without effecting the original directory + creating the jfrog config for artifactory resolution
 	tmpWd, err := fileutils.CreateTempDir()
 	if err != nil {
-		err = fmt.Errorf("failed creating temporary dir: %w", err)
```
omerzi, line 63: '%w' prints the string out of the error, it's better to use `%w`.

xray/commands/audit/sca/nuget/nuget.go
```diff
@@ -40,7 +40,7 @@ func BuildDependencyTree(params utils.AuditParams) (dependencyTree []*xrayUtils.
 	}
 	sol, err := solution.Load(wd, "", log.Logger)
 	if err != nil {
```
omerzi, line 42:
```suggestion
	if err != nil &&   !strings.Contains(err.Error(), "could not find global packages path at:" {
		// If an error arises due to the inability to locate the global package path, we aim to proceed, as this issue is expected to be resolved upon project restoration
		return
	}
```

Additionally, assign the expected error to a constant with a clear name that represents the error we intend to handle, and explain further in the comment why we didn't fail when this error returned.

omerzi (changes requested):

PR #1070, opened 2023-12-11
Show sha256 for Xray's DBSyncV3 files to be downloaded

eranbrodet:


eranbrodet:

xray/commands/offlineupdate/offlineupdate.go
```diff
@@ -110,11 +113,27 @@ func getURLsToDownloadDBSyncV3(responseBody []byte, isPeriodicUpdate bool) ([]st
 		}
 		for _, packageUrl := range onboardingResponse {
 			urlsToDownload = append(urlsToDownload, packageUrl.DownloadUrl)
+			logV3UpdateFileChecksum(packageUrl)
 		}
 	}
 	return urlsToDownload, nil
 }
 
+func logV3UpdateFileChecksum(updateFileItem V3UpdateResponseItem) {
```
eyalbe4, line 122: Let's add a unit test for this function in `offlineupdate_test.go`. Let's follow the table-testing pattern. See other tests in that file as a reference.

eyalbe4 (changes requested):

xray/commands/offlineupdate/offlineupdate.go
eranbrodet: Done

xray/commands/offlineupdate/offlineupdate_test.go
```diff
@@ -121,3 +121,17 @@ func TestGetValidStreamsString(t *testing.T) {
...
+	tests := []struct {
+		item V3UpdateResponseItem
+	}{
+		{item: V3UpdateResponseItem{DownloadUrl: "http://www.example.com/some/path/filename.tar.zst", Sha256: "0000000000000000000000000000000000000000000000000000000000000000"}},
+		{item: V3UpdateResponseItem{DownloadUrl: ">bad_url%+o", Sha256: "0000000000000000000000000000000000000000000000000000000000000000"}},
+		{item: V3UpdateResponseItem{DownloadUrl: "http://www.example.com/some/path/filename.tar.zst"}},
+	}
+
+	for _, test := range tests {
+		logV3UpdateFileChecksum(test.item)
```
eyalbe4, line 135: It appears that this test doesn't asset anything. It will therefore always pass.
Herre's one way to really test the functionality - 
You can replace the existing `logV3UpdateFileChecksum` function into two functions with a new function named `generateV3UpdateFileChecksumMessage`. The new function will return a string, and therefire, ut'll be really easy to assert its returned value inside this test function. You code can then use this function as follows - 
```
log.Info(generateV3UpdateFileChecksumMessage())
```

eyalbe4 (changes requested):

xray/commands/offlineupdate/offlineupdate_test.go
eranbrodet: I agree the test isn't interesting, that's why I didn't write one to begin with, all the function does is log something without having any external effect and therefore nothing interesting to test. The only sort of interesting logic is a fallback for a url that doesn't parse, but we don't really care what gets logged in that case, if it ever happens that the url is bad, the log is not that interesting because the attempt to download the file from that url will fail (and the full url is logged when downloading, so it's not that we need the information). 
When you asked for a test I complied, thinking you might be chasing after coverage, and wrote something that goes through the different flows, but it's a bit pointless.

So the question now is, why write extra logic into the code and split it up to write more functions, just for the test that we didn't need? The new test wouldn't be really testing anything interesting either.

If you insist that this is a good idea, I will add the test as you describe, but in my opinion it will still be a badly written/needless test.
Also I was debating with myself just logging the full url always and not having even that possible "error" case, and decided to parse the url just to print something a little more aesthetic, which we don't really have to do. So if you prefer, I don't mind not parsing the url and just logging the full url always and not having any logic that would justify a test at all.

Let me know what you prefer.
Thanks.

eyalbe4 (changes requested):
* We're adding tests to ensure the code works as expected, regardless of whether the functionality has an external affect or not. 
* Testing this porperly is super easy. No extra logic what so ever. Simply have a function that returns the log output as a string, and then test its output. The tests table should include all cases.

eranbrodet:
> * We're adding tests to ensure the code works as expected, regardless of whether the functionality has an external affect or not.
> 
>     * Testing this porperly is super easy. No extra logic what so ever. Simply have a function that returns the log output as a string, and then test its output. The tests table should include all cases.

I've extracted getting the filename string to a different function and changed the test to check that returned string.
Works for you?
I think adding a function that is "get string for specific log" is a bad code-smell, at least parsing the filename is reasonably something that justify being its own function and we can put in the test you desire.

PR #1088, opened 2023-12-25
Transfer - Attempt to transfer ".jfrog" directory in Docker repositories

yahavi:
-----

Certain files, locally generated in Artifactory, shouldn't be transferred.
The `.jfrog` directory, unintentionally included in Docker repository transfers, is triggering errors documented in a CSV file:
```csv
docker-local,.jfrog,,0,false,0,false,FAIL,404,,2023-12-21T16:31:01Z
```

The central cause of the transfer issues stemmed from an incorrect path specification within the locally generated API call. The path `./.jfrog` was inadvertently used, whereas the accurate path should have been solely `.jfrog`. This discrepancy led to the errors encountered.


artifactory/commands/transferfiles/localgeneratedfilter_test.go
```diff
@@ -99,3 +100,20 @@ func TestFilterLocallyGeneratedEnabled(t *testing.T) {
 	assert.False(t, NewLocallyGenerated(context.Background(), servicesManager, "7.54.5").IsEnabled())
 	assert.False(t, NewLocallyGenerated(context.Background(), servicesManager, "6.0.0").IsEnabled())
 }
+
+var getPathInRepoCases = []struct {
```
eyalbe4, line 104: This variable can be made a local variable inside the `TestGetPathInRepo` function below.

eyalbe4 (changes requested):

artifactory/commands/transferfiles/localgeneratedfilter_test.go
yahavi: We agreed to keep it outside the function to make the test short and clear

PR #1089, opened 2023-12-25
Fix memory & time explosion during dependency tree construction for java based projects

eranturgeman:
-----

This PR fixes the bug of memory & time explosion while constructing Gradle/ Maven dependencies tree in big projects
A limitation to the number of impact paths per dependency was set to 10 (per project's module)

### DIFF
--- a/xray/commands/audit/sca/java/deptreemanager.go
+++ b/xray/commands/audit/sca/java/deptreemanager.go
@@ -90,23 +90,6 @@ func getGraphFromDepTree(outputFilePaths string) (depsGraph []*xrayUtils.GraphNo
 	return
 }
 
-func populateDependencyTree(currNode *xrayUtils.GraphNode, currNodeId string, moduleTree *moduleDepTree, uniqueDepsSet *datastructures.Set[string]) {
-	if currNode.NodeHasLoop() {
-		return
-	}
-	for _, childId := range moduleTree.Nodes[currNodeId].Children {
-		childGav := GavPackageTypeIdentifier + childId
-		childNode := &xrayUtils.GraphNode{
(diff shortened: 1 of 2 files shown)


xray/commands/audit/sca/java/deptreemanager.go
```diff
@@ -58,45 +59,39 @@ type depTreeNode struct {
...
+			dependencyId := GavPackageTypeIdentifier + depName
+			var childrenList []string
+			for _, childName := range dependency.Children {
+				childId := GavPackageTypeIdentifier + childName
+				childrenList = append(childrenList, childId)
+			}
+			if len(childrenList) > 0 {
+				moduleTreeMap[dependencyId] = childrenList
+			}
+		}
```
omerzi, line 84: Let's extract this code into a new function, so it will be more readable.

xray/commands/audit/sca/java/deptreemanager.go
```diff
@@ -58,45 +59,39 @@ type depTreeNode struct {
...
+		moduleTree, moduleUniqueDeps := sca.BuildXrayDependencyTree(moduleTreeMap, GavPackageTypeIdentifier+module.Root)
+		depsGraph = append(depsGraph, moduleTree)
+		for _, depToAdd := range moduleUniqueDeps {
+			uniqueDepsSet.Add(depToAdd)
 		}
```
omerzi, line 89: The Set in this case becomes irrelevant after your changes because we already obtain 'uniqueDeps' as a slice of strings. Therefore, we can skip initializing a set and directly convert it to a slice of strings (line 91).

omerzi (changes requested):

PR #1116, opened 2024-01-25
Optimization to clearMavenDepTreeRun call

eranturgeman:
-----

### DIFF
--- a/utils/java/mvn.go
+++ b/utils/java/mvn.go
@@ -83,9 +83,6 @@ func (mdt *MavenDepTreeManager) RunMavenDepTree() (depTreeOutput string, clearMa
 	// depTreeExecDir is a temp directory for all the files that are required for the maven-dep-tree run
 	depTreeExecDir, clearMavenDepTreeRun, err := mdt.CreateTempDirWithSettingsXmlIfNeeded()
 	if err != nil {
-		if clearMavenDepTreeRun != nil {
-			err = errors.Join(err, clearMavenDepTreeRun())
-		}
 		return
 	}
 	if err = mdt.installMavenDepTreePlugin(depTreeExecDir); err != nil {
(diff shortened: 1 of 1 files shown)


utils/java/mvn.go
```diff
@@ -83,9 +83,6 @@ func (mdt *MavenDepTreeManager) RunMavenDepTree() (depTreeOutput string, clearMa
 	// depTreeExecDir is a temp directory for all the files that are required for the maven-dep-tree run
 	depTreeExecDir, clearMavenDepTreeRun, err := mdt.CreateTempDirWithSettingsXmlIfNeeded()
```
attiasas, line 84: can CreateTempDirWithSettingsXmlIfNeeded be private? I can see its only usage at core is at mvn.go?

eranturgeman: it is also used in mavenPackageHandler to avoid the execution of the entire plugin

utils/java/mvn.go
```diff
@@ -231,5 +228,9 @@ func (mdt *MavenDepTreeManager) CreateTempDirWithSettingsXmlIfNeeded() (tempDirP
 	if mdt.depsRepo != "" {
 		err = mdt.createSettingsXmlWithConfiguredArtifactory(tempDirPath)
 	}
+	if err != nil {
+		err = errors.Join(err, clearMavenDepTreeRun())
+		clearMavenDepTreeRun = nil
+	}
```
attiasas, line 234: Now that the clean, in the case of error, occurs inside the `CreateTempDirWithSettingsXmlIfNeeded`
you need to also take care of the code that uses it...
`RunMavenDepTree:86` needs to be removed.
also I can see a call for `RunMavenDepTree` in case of error at: `buildMavenDependencyTree:66`
Make sure to take care of all the cases for this

attiasas (changes requested):
add description

PR #1118, opened 2024-01-25
Maven support for curation audit cmd.

asafambar:
-----

### DIFF
--- a/utils/java/mvn.go
+++ b/utils/java/mvn.go
@@ -22,7 +22,7 @@ const (
 	mavenDepTreeJarFile    = "maven-dep-tree.jar"
 	mavenDepTreeOutputFile = "mavendeptree.out"
 	// Changing this version also requires a change in MAVEN_DEP_TREE_VERSION within buildscripts/download_jars.sh
-	mavenDepTreeVersion = "1.0.3"
+	mavenDepTreeVersion = "1.1.0"
 	settingsXmlFile     = "settings.xml"
 )
 
@@ -43,8 +43,10 @@ var mavenDepTreeJar []byte
 
 type MavenDepTreeManager struct {
 	DepTreeManager
-	isInstalled         bool
-	isCurationCmd       bool
+	isInstalled bool
+	// this flag its curation command, it will set dedicated cache and download url.
+	isCurationCmd bool
+	// path to the curation dedicated cache
 	curationCacheFolder string
 	cmdName             MavenDepTreeCmd
 	settingsXmlPath     string
--- a/utils/java/mvn.go
+++ b/utils/java/mvn.go
@@ -22,7 +22,7 @@ const (
 	mavenDepTreeJarFile    = "maven-dep-tree.jar"
 	mavenDepTreeOutputFile = "mavendeptree.out"
 	// Changing this version also requires a change in MAVEN_DEP_TREE_VERSION within buildscripts/download_jars.sh
-	mavenDepTreeVersion = "1.0.2"
+	mavenDepTreeVersion = "1.0.3"
 	settingsXmlFile     = "settings.xml"
 )
 
--- a/utils/java/mvn.go
+++ b/utils/java/mvn.go
@@ -7,6 +7,7 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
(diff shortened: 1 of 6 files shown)


utils/java/mvn.go
```diff
@@ -21,7 +22,7 @@ const (
 	mavenDepTreeJarFile    = "maven-dep-tree.jar"
 	mavenDepTreeOutputFile = "mavendeptree.out"
 	// Changing this version also requires a change in MAVEN_DEP_TREE_VERSION within buildscripts/download_jars.sh
-	mavenDepTreeVersion = "1.0.2"
+	mavenDepTreeVersion = "1.0.3"
```
attiasas, line 25:
```suggestion
	mavenDepTreeVersion = "1.1.0"
```
latest version is 1.1.0, is it require to be 1.0.3? don't forget to run download_jars.sh if changing.

utils/java/mvn.go
```diff
@@ -42,37 +43,39 @@ var mavenDepTreeJar []byte
 
 type MavenDepTreeManager struct {
 	DepTreeManager
-	isInstalled     bool
-	cmdName         MavenDepTreeCmd
-	settingsXmlPath string
+	isInstalled         bool
+	isCurationCmd       bool
+	curationCacheFolder string
```
attiasas, line 48:
```suggestion
	isCurationCmd       bool
	curationCacheFolder string
```
Please add a comment above to describe why and what are each attribute

utils/java/mvn.go
```diff
@@ -123,6 +126,9 @@ func (mdt *MavenDepTreeManager) execMavenDepTree(depTreeExecDir string) (string,
 func (mdt *MavenDepTreeManager) runTreeCmd(depTreeExecDir string) (string, error) {
 	mavenDepTreePath := filepath.Join(depTreeExecDir, mavenDepTreeOutputFile)
 	goals := []string{"com.jfrog:maven-dep-tree:" + mavenDepTreeVersion + ":" + string(Tree), "-DdepsTreeOutputFile=" + mavenDepTreePath, "-B"}
+	if mdt.isCurationCmd {
+		goals = append(goals, "-Dmaven.repo.local="+mdt.curationCacheFolder)
```
attiasas, line 130:
```suggestion
	if mdt.isCurationCmd && mdt.curationCacheFolder != "" {
		goals = append(goals, "-Dmaven.repo.local="+mdt.curationCacheFolder)
```
what happend if we are setting it to an empty path? can this

attiasas (changes requested):
Please add a test that uses the new attributes (curation and cache)

utils/java/mvn.go
asafambar: I guess 1.1.0 will work as well, what is download_jars.sh ?

utils/java/mvn.go
asafambar: well its possible to do that from an enviroment variable, thi option by default go to the path under the home folder, If user explicitly set it to be empty then it will go to the default cahce of maven which is under ~/.m2/repository. we set it to be under a different cache folder so the curtion packages downloaded during the build of tree wont infect the default cache(some of them might be blocked by curation policies) so ist kind of not recommended setup.

utils/java/mvn.go
attiasas: run locally https://github.com/jfrog/jfrog-cli-core/blob/master/buildscripts/download-jars.sh

utils/java/mvn.go
attiasas: just making sure you are taking into account this case.

utils/java/mvn.go
```diff
@@ -21,7 +22,7 @@ const (
 	mavenDepTreeJarFile    = "maven-dep-tree.jar"
 	mavenDepTreeOutputFile = "mavendeptree.out"
 	// Changing this version also requires a change in MAVEN_DEP_TREE_VERSION within buildscripts/download_jars.sh
```
yahavi, line 24: Please also update the value in `buildscripts/download_jars.sh` and execute the script (run `./buildscripts/download_jars.sh`) to update the jar.

PR #1143, opened 2024-02-29
Artifactory Release Lifecycle Management  - Add Export bundle function

EyalDelarea:
-----

depends on https://github.com/jfrog/jfrog-client-go/pull/917


lifecycle/export.go
```diff
@@ -0,0 +1,149 @@
...
+	clientConfig "github.com/jfrog/jfrog-client-go/config"
+	"github.com/jfrog/jfrog-client-go/lifecycle/services"
+	utils2 "github.com/jfrog/jfrog-client-go/utils"
+	"github.com/jfrog/jfrog-client-go/utils/log"
+	"strings"
+)
+
+type ReleaseBundleExportCommand struct {
+	releaseBundleCmd
+	modifications          *services.Modifications
```
RobiNino, line 20: Let's avoid using pointers if they are not necessary.

lifecycle/export.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func (rbe *ReleaseBundleExportCommand) Run() (err error) {
+	if err = validateArtifactoryVersionSupported(rbe.serverDetails); err != nil {
+		return
+	}
+	servicesManager, rbDetails, queryParams, err := rbe.getPrerequisites()
+	if err != nil {
+		log.Debug("Failed getting prerequisites for exporting command, error: ", err.Error())
+		return
```
RobiNino, line 31: Debug logs can be missed. If you want to add a message to the error, use:

```suggestion
		return errorutils.CheckErrorf("failed getting prerequisites for exporting command, error: '%s'", err.Error())
```

lifecycle/export.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	// Start the Export process and wait for completion
+	log.Info("Exporting Release Bundle archive...")
+	releaseBundleExportParams := NewReleaseBundleExportParams(rbDetails, *rbe.modifications)
+	exportResponse, err := servicesManager.ExportReleaseBundle(releaseBundleExportParams, queryParams)
+	if err != nil {
+		log.Debug("Failed exporting release bundle, error: ", err.Error())
```
RobiNino, line 38: Same as previous comment

lifecycle/export.go
```diff
@@ -0,0 +1,149 @@
...
+		ValidateSymlink: downloadConfiguration.ValidateSymlink,
+		MinSplitSize:    downloadConfiguration.MinSplitSize,
+		SplitCount:      downloadConfiguration.SplitCount,
+		SkipChecksum:    downloadConfiguration.SkipChecksum,
+	}
+	artifactoryServiceManager, err := createArtifactoryServiceManager(rbe.serverDetails)
+	if err != nil {
+		return
+	}
+	sum, err := artifactoryServiceManager.DownloadFilesWithSummary(downloadParams)
```
RobiNino, line 75: Opt to use `artifactoryServiceManager.DownloadFiles` here since you don't use the summary

lifecycle/export.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	log.Output(utils2.IndentJson(content))
+	return
+}
+
+// Download the exported release bundle using artifactory service manager
+func (rbe *ReleaseBundleExportCommand) downloadReleaseBundle(exportResponse services.ReleaseBundleExportedStatusResponse, downloadConfiguration *artUtils.DownloadConfiguration) (cleanUp func() error, err error) {
+	downloadParams := artServices.DownloadParams{
+		CommonParams: &utils.CommonParams{
+			Pattern: strings.TrimPrefix(exportResponse.RelativeUrl, "/"),
```
RobiNino, line 63: Use `Target` here to specify a target download location

lifecycle/export.go
```diff
@@ -0,0 +1,149 @@
...
+		},
+		Symlink:         downloadConfiguration.Symlink,
```
RobiNino, line 65: The only params relevant here might be `MinSplitSize` and `SplitCount`. The rest can be removed

lifecycle/export.go
```diff
@@ -0,0 +1,149 @@
...
+	log.Debug("Downloading the exported bundle...")
+	cleanUp, err := rbe.downloadReleaseBundle(exportResponse, rbe.downloadConfigurations)
+	defer func() {
+		err = cleanUp()
+	}()
+	if err != nil {
+		return
+	}
+	log.Info("Successfully Downloaded Release Bundle archive")
+	content, err := json.Marshal(exportResponse)
```
RobiNino, line 51: I'm not sure there's an added value to logging the response if we download the file.

lifecycle/export.go
```diff
@@ -0,0 +1,149 @@
...
+		CommonParams: &utils.CommonParams{
+			Pattern: strings.TrimPrefix(exportResponse.RelativeUrl, "/"),
+		},
+		Symlink:         downloadConfiguration.Symlink,
+		ValidateSymlink: downloadConfiguration.ValidateSymlink,
+		MinSplitSize:    downloadConfiguration.MinSplitSize,
+		SplitCount:      downloadConfiguration.SplitCount,
+		SkipChecksum:    downloadConfiguration.SkipChecksum,
+	}
+	artifactoryServiceManager, err := createArtifactoryServiceManager(rbe.serverDetails)
```
RobiNino, line 71: In order for the progress bar to work, we need to initialize a service manager with progress (see `CreateDownloadServiceManager` for example). This function doesn't do it.
Let's be sure we want to add this at this stage of time. If we do, it will require a bit more configuration both here and on the CLI side.

RobiNino (changes requested):

lifecycle/export.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	// Start the Export process and wait for completion
+	log.Info("Exporting Release Bundle archive...")
+	exportResponse, err := servicesManager.ExportReleaseBundle(rbDetails, rbe.modifications, queryParams)
+	if err != nil {
+		return errorutils.CheckErrorf("Failed exporting release bundle, error: '%s'", err.Error())
+	}
+	// Download the exported bundle
+	log.Debug("Downloading the exported bundle...")
+	_, failed, err := rbe.downloadReleaseBundle(exportResponse, rbe.downloadConfigurations)
```
RobiNino, line 40: Might want not to ignore the `downloaded int` to assert the release bundle was truly downloaded.

lifecycle/export.go
```diff
@@ -0,0 +1,133 @@
...
+func (rbe *ReleaseBundleExportCommand) CommandName() string {
+	return "rb_export"
+}
+
+func NewReleaseBundleExportCommand() *ReleaseBundleExportCommand {
+	return &ReleaseBundleExportCommand{}
+}
+func (rbe *ReleaseBundleExportCommand) SetServerDetails(serverDetails *config.ServerDetails) *ReleaseBundleExportCommand {
+	rbe.serverDetails = serverDetails
+	rbe.releaseBundleCmd.serverDetails = serverDetails
```
RobiNino, line 78: These two lines are a duplication. You can remove the second one.

lifecycle/export.go
```diff
@@ -0,0 +1,133 @@
...
+	rbe.rbProjectKey = project
+	return rbe
+}
+
+func (rbe *ReleaseBundleExportCommand) SetDownloadConfiguration(downloadConfig artUtils.DownloadConfiguration) *ReleaseBundleExportCommand {
+	rbe.downloadConfigurations = downloadConfig
+	return rbe
+}
+
+func (rbe *ReleaseBundleExportCommand) SetTargetPath(target string) *ReleaseBundleExportCommand {
```
RobiNino, line 106: We can allow target path to include a file name to, not necessarily a directory. Lets mimic the behavior as in the download command.

RobiNino (changes requested):

lifecycle/export.go
EyalDelarea: As for this, I've tested with a 1GB bundle at the logs seems pretty well.
I guess if we will see the need in the future we can implement a full progress bar.

PR #1149, opened 2024-03-06
Publish Multiple npm packages (workspaces)

EyalDelarea:
-----

### DIFF
--- a/artifactory/commands/npm/publish.go
+++ b/artifactory/commands/npm/publish.go
@@ -50,14 +50,14 @@ type NpmPublishCommandArgs struct {
 type NpmPublishCommand struct {
 	configFilePath  string
 	commandName     string
-	results         []*commandsutils.Result
+	result          *commandsutils.Result
 	detailedSummary bool
 	npmVersion      *version.Version
 	*NpmPublishCommandArgs
 }
 
 func NewNpmPublishCommand() *NpmPublishCommand {
-	return &NpmPublishCommand{NpmPublishCommandArgs: NewNpmPublishCommandArgs(), commandName: "rt_npm_publish", results: []*commandsutils.Result{}}
+	return &NpmPublishCommand{NpmPublishCommandArgs: NewNpmPublishCommandArgs(), commandName: "rt_npm_publish", result: new(commandsutils.Result)}
 }
 
 func NewNpmPublishCommandArgs() *NpmPublishCommandArgs {
@@ -97,8 +97,8 @@ func (npc *NpmPublishCommand) SetScanOutputFormat(format format.OutputFormat) *N
 	return npc
 }
 
-func (npc *NpmPublishCommand) Result() []*commandsutils.Result {
-	return npc.results
+func (npc *NpmPublishCommand) Result() *commandsutils.Result {
+	return npc.result
 }
 
 func (npc *NpmPublishCommand) IsDetailedSummary() bool {
@@ -219,7 +219,6 @@ func (npc *NpmPublishCommand) CommandName() string {
 
 func (npc *NpmPublishCommand) preparePrerequisites() error {
 	npc.packedFilesPath = make([]string, 0)
-	npc.results = make([]*commandsutils.Result, 0)
 	currentDir, err := os.Getwd()
 	if err != nil {
 		return errorutils.CheckError(err)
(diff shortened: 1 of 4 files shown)


artifactory/commands/npm/publish.go
```diff
@@ -37,7 +38,7 @@ type NpmPublishCommandArgs struct {
 	executablePath         string
 	workingDirectory       string
 	collectBuildInfo       bool
-	packedFilePath         string
+	packedFilesPath        []string
```
sverdlov93, line 41:
```suggestion
	packedFilePaths        []string
```

artifactory/utils/npm/pack.go
```diff
@@ -27,8 +27,7 @@ func createPackCmdConfig(executablePath string, splitFlags []string) *npmutils.N
 	}
 }
 
-func getPackageFileNameFromOutput(output string) (string, error) {
+func getPackageFileNameFromOutput(output string) []string {
 	output = strings.TrimSpace(output)
-	lines := strings.Split(output, "\n")
-	return strings.TrimSpace(lines[len(lines)-1]), nil
+	return strings.Split(output, "\n")
```
sverdlov93, line 32: will work on windows?

artifactory/commands/npm/publish.go
```diff
@@ -436,18 +458,20 @@ func (npc *NpmPublishCommand) readPackageInfoFromTarball() (err error) {
 	}
 }
 
-func deleteCreatedTarballAndError(packedFilePath string, currentError error) error {
-	if err := deleteCreatedTarball(packedFilePath); err != nil {
+func deleteCreatedTarballAndError(packedFilesPath []string, currentError error) error {
+	if err := deleteCreatedTarball(packedFilesPath); err != nil {
```
sverdlov93, line 462: lets replace this shitty func with a simple errors.join

artifactory/commands/npm/publish.go
```diff
@@ -261,8 +263,11 @@ func (npc *NpmPublishCommand) pack() error {
 		return err
 	}
 
-	npc.packedFilePath = filepath.Join(tarballDir, packageFileName)
-	log.Debug("Created npm package at", npc.packedFilePath)
+	for index, packageFileName := range packedFileNames {
+		npc.packedFilesPath = append(npc.packedFilesPath, filepath.Join(tarballDir, packageFileName))
+		log.Debug("Created npm package at", npc.packedFilesPath[index])
+	}
```
sverdlov93, line 269:
```suggestion
	for _, packageFileName := range packedFileNames {
		npc.packedFilesPath = append(npc.packedFilesPath, filepath.Join(tarballDir, packageFileName))
	}
```
nothing is created here so it's bit odd log

artifactory/commands/npm/publish.go
```diff
@@ -341,9 +348,15 @@ func (npc *NpmPublishCommand) doDeploy(target string, artDetails *config.ServerD
...
-			npc.result.SetSuccessCount(summary.TotalSucceeded)
+			npc.result.SetFailCount(npc.result.FailCount() + totalFailed)
+			npc.result.SetSuccessCount(npc.result.SuccessCount() + summary.TotalSucceeded)
+			if npc.result.Reader() == nil {
+				npc.result.SetReader(summary.TransferDetailsReader)
+			} else {
+				if err = npc.appendReader(summary); err != nil {
+					return err
+				}
+			}
```
sverdlov93, line 359: too much if else if else if else in this func, can we simplify it?

sverdlov93 (commented):
Great work! lets add a test to JFrog CLI with --workspaces flag and moe then one pack

artifactory/utils/npm/pack_test.go
EyalDelarea, line 1: I've removed this test as the test case was irrelevant.
We checked the whole output of the CMD and tried to extract the tar.gz file name, as in practice, the npm cmd already gives us the name. and all we have to do now is to split the file names if we have more then one.
I've deleted the test here, and added a full test in CLI side to publish multiple packages.

artifactory/utils/npm/pack.go
EyalDelarea: Looks like from the CLI test that it is working on windows 👍

PR #1153, opened 2024-03-11
Artifactory Release Lifecycle Management - Add Import bundle function

EyalDelarea:
-----

### DIFF
--- a/lifecycle/import.go
+++ b/lifecycle/import.go
@@ -1,0 +1,73 @@
+package lifecycle
+
+import (
+	"github.com/jfrog/jfrog-cli-core/v2/utils/config"
+	"github.com/jfrog/jfrog-cli-core/v2/utils/coreutils"
+	"github.com/jfrog/jfrog-client-go/artifactory"
+	clientConfig "github.com/jfrog/jfrog-client-go/config"
+	"github.com/jfrog/jfrog-client-go/utils/log"
+)
+
+type ReleaseBundleImportCommand struct {
+	releaseBundleCmd
+	filePath string
+}
+
+func (rbi *ReleaseBundleImportCommand) ServerDetails() (*config.ServerDetails, error) {
+	return rbi.serverDetails, nil
+}
+
+func (rbi *ReleaseBundleImportCommand) CommandName() string {
+	return "rb_export"
+}
+
+func NewReleaseBundleImportCommand() *ReleaseBundleImportCommand {
+	return &ReleaseBundleImportCommand{}
+}
+func (rbi *ReleaseBundleImportCommand) SetServerDetails(serverDetails *config.ServerDetails) *ReleaseBundleImportCommand {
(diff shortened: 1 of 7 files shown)


lifecycle/import.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (rbi *ReleaseBundleImportCommand) Run() (err error) {
+	if err = validateArtifactoryVersionSupported(rbi.serverDetails); err != nil {
+		return
+	}
+	artService, err := createArtifactoryServiceManager(rbi.serverDetails)
+	if err != nil {
+		return
+	}
+	log.Info("Importing Release Bundle Archive...")
```
RobiNino, line 43:
```suggestion
	log.Info("Importing the release bundle archive...")
```

lifecycle/import.go
```diff
@@ -0,0 +1,49 @@
...
+	if err = artService.ReleaseBundleImport(rbi.filePath); err != nil {
+		return
+	}
+	log.Info("Successfully Imported Release Bundle archive")
```
RobiNino, line 47:
```suggestion
	log.Info("Successfully imported release bundle")
```

lifecycle/import.go
```diff
@@ -0,0 +1,49 @@
...
+func (rbi *ReleaseBundleImportCommand) Run() (err error) {
+	if err = validateArtifactoryVersionSupported(rbi.serverDetails); err != nil {
+		return
+	}
+	artService, err := createArtifactoryServiceManager(rbi.serverDetails)
+	if err != nil {
+		return
+	}
+	log.Info("Importing Release Bundle Archive...")
+	if err = artService.ReleaseBundleImport(rbi.filePath); err != nil {
```
RobiNino, line 44: Lets check if the file exists before trying to import it.

RobiNino (changes requested):

lifecycle/import.go
```diff
@@ -0,0 +1,49 @@
...
+	return rbi
+}
+
+func (rbi *ReleaseBundleImportCommand) SetFilepath(filePath string) *ReleaseBundleImportCommand {
+	rbi.filePath = filePath
+	return rbi
+}
+
+func (rbi *ReleaseBundleImportCommand) Run() (err error) {
+	if err = validateArtifactoryVersionSupported(rbi.serverDetails); err != nil {
```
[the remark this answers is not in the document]
EyalDelarea: i've added a function that will check if it is cloud platform, but what do you suggest about the version constraint? there is guarantee that it will be fixed the next patch.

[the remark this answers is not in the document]
RobiNino: The problem with checking whether the instance is a cloud is that it requires admin privileges, which is not a constraint we would like to set.
To be on the safe side, I would suggest to keep the command hidden till the patch is officially out, then exposing it with the same minimal version for both cloud and self hosted.

PR #1154, opened 2024-03-12
Remove applicabilityScannable attribute for technology

attiasas:
-----
This attribute is no longer needed. the new scanners will return a new status 'NOT_SCANNED' if not supported yet.


hadarshjfrog:
The code change itself looks good - my only concern - is how can we validate that when scanning a non-supported technology - just nothing well happened (as expected) - and not break anything?

PR #1162, opened 2024-03-26
Fix npm publish with provided tarball

EyalDelarea:
-----

# Fix npm pack when reading a pre packed npm package
When reading for an already packed npm package, there was a missing file path that wasn't set because of the changes to handle workspaces.
This PR should fix it by setting the file path.


eyalbe4 (changes requested):
Let's add an integration test in the jfrog-cli project, that also tests the missing usage this PR fixes.

PR #1165, opened 2024-03-31
New XSC analytics metrics capabilities

gailazar300:
-----

1. Add XscAuthConfig.
2. Add XSCmock server.
3. Add technology to language mapping.

depends on: https://github.com/jfrog/jfrog-client-go/pull/928


utils/config/config.go
```diff
@@ -708,6 +709,19 @@ func (serverDetails *ServerDetails) CreateXrayAuthConfig() (auth.ServiceDetails,
 	return serverDetails.createAuthConfig(artAuth)
 }
 
+func (serverDetails *ServerDetails) CreateXscAuthConfig() (auth.ServiceDetails, error) {
+	ascAuth := xscAuth.NewXscDetails()
+	ascAuth.SetUrl(convertXrayUrlToXscUrl(serverDetails.XrayUrl))
```
attiasas, line 714:
```suggestion
	ascAuth.SetUrl(serverDetails.getXscUrlFromXrayUrl())
```
private method, called once, passing an argument that is also an attribute.
should use different implementation

utils/config/config.go
```diff
@@ -708,6 +709,19 @@ func (serverDetails *ServerDetails) CreateXrayAuthConfig() (auth.ServiceDetails,
...
+	return serverDetails.createAuthConfig(ascAuth)
+}
+
+// Xray ans Xsc will always have the same platform url.
```
attiasas, line 718:
```suggestion
// Xray and Xsc will always have the same platform url.
```

utils/config/config.go
```diff
@@ -708,6 +709,19 @@ func (serverDetails *ServerDetails) CreateXrayAuthConfig() (auth.ServiceDetails,
...
+func convertXrayUrlToXscUrl(xrayUrl string) string {
+	xscUrl := strings.TrimSuffix(xrayUrl, "/")
+	xscUrl = strings.TrimSuffix(xscUrl, "xray")
+	return xscUrl + "xsc/"
+}
```
attiasas, line 723:
```suggestion
func (serverDetails *ServerDetails) getXscUrlFromXrayUrl() string {
	xscUrl := strings.TrimSuffix(serverDetails.XrayUrl, "/")
	xscUrl = strings.TrimSuffix(xscUrl, "xray")
	return xscUrl + "xsc/"
}
```

utils/coreutils/techutils.go
```diff
@@ -29,13 +29,30 @@ const (
 	Go     Technology = "go"
 	Pip    Technology = "pip"
 	Pipenv Technology = "pipenv"
+	Pypi   Technology = "pypi"
```
attiasas, line 32: Pypi is not a technology.
we have 3 tools for python: `Pip`, `Pipenv` and `Poetry`
```suggestion
	Pypi = "pypi"
```

utils/coreutils/techutils.go
```diff
@@ -29,13 +29,30 @@ const (
...
+		Go:     "Go",
+		Maven:  "Java",
+		Gradle: "Java",
+		Nuget:  "C#",
+		Dotnet: "C#",
+		Yarn:   "JavaScript",
+		Pnpm:   "JavaScript",
+	}
+	return languageMap[technology]
+}
```
attiasas, line 55:
```suggestion
type CodeLanguage string

const (
    JavaScript CodeLanguage = "javascript"
    Python     CodeLanguage = "python"
    GoLang     CodeLanguage = "go"
    Java       CodeLanguage = "java"
    CSharp     CodeLanguage = "C#"

)
```
and add `Language` attribute to the `TechData` objects

attiasas (changes requested):
Update the PR name to fit the changes better.
Add at the description the changes that were made.

attiasas (changes requested):
Add tests to check connection of language and tech at `techutils_test.go`

common/tests/utils.go
```diff
@@ -76,6 +76,15 @@ func CreateDsRestsMockServer(t *testing.T, testHandler restsTestHandler) (*httpt
...
 }
 
+func CreateXscRestsMockServer(t *testing.T, testHandler restsTestHandler) (*httptest.Server, *config.ServerDetails, artifactory.ArtifactoryServicesManager) {
+	testServer := CreateRestsMockServer(testHandler)
+	serverDetails := &config.ServerDetails{Url: testServer.URL + "/", XrayUrl: testServer.URL + "/xray/"}
+
+	serviceManager, err := utils.CreateServiceManager(serverDetails, -1, 0, false)
+	assert.NoError(t, err)
+	return testServer, serverDetails, serviceManager
+}
```
attiasas, line 86: This can be created at: https://github.com/jfrog/jfrog-cli-security/pull/47
where it is used. We probably will only use mocks for Xray in our repository
added to `tests/utils/test_utils.go` / `tests/utils/test_config.go`

utils/config/config.go
```diff
@@ -572,6 +573,7 @@ type ServerDetails struct {
 	ArtifactoryUrl                  string `json:"artifactoryUrl,omitempty"`
 	DistributionUrl                 string `json:"distributionUrl,omitempty"`
 	XrayUrl                         string `json:"xrayUrl,omitempty"`
+	XscUrl                          string `json:"xscUrl,omitempty"`
```
attiasas, line 576: Why did you add that if it is never used?
at client-go you are using GetUrl() and you are setings using the value at `XrayUrl`

utils/coreutils/techutils.go
```diff
@@ -37,6 +37,33 @@ const (
...
+		Go:     GoLang,
+		Maven:  Java,
+		Gradle: Java,
+		Nuget:  CSharp,
+		Dotnet: CSharp,
+		Yarn:   JavaScript,
+		Pnpm:   JavaScript,
+	}
+	return languageMap[technology]
+}
```
attiasas, line 55: Instead of this func, you can attach the value directly to the `TechData` as new attribute `Language CodeLanguage`

attiasas (changes requested):

utils/config/config.go
gailazar300: we use this field for logic in security-cli

utils/coreutils/techutils.go
gailazar300: I don't understand your comment, but what is wrong with this implementation?

common/tests/utils.go
gailazar300: all other mocks are there, so it makes sense that they will all be together in the same place.

eyalbe4 (changes requested):
The code introduced in this PR should be added to the `jfrog-cli-security` project.

PR #1171, opened 2024-04-14
Fix npm pack command with prePack script

EyalDelarea:
-----

modified the `getPackageFileNameFromOutput` function to extract .tgz files from the output, which as we saw, can be different when pre pack scripts exists or any other parameters which can effect the output of the npm pack command

### DIFF
--- a/artifactory/utils/npm/pack.go
+++ b/artifactory/utils/npm/pack.go
@@ -27,7 +27,20 @@ func createPackCmdConfig(executablePath string, splitFlags []string) *npmutils.N
 	}
 }
 
+// Extracts packed file names from npm pack command output
+// The output can differ when a prePack script exists,
+// This function will filter the output and search for the .tgz files.
 func getPackageFileNameFromOutput(output string) []string {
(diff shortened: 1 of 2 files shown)


RobiNino (approved):
This generally LGTM. 
The only problem with this approach is that we might unintentionally identify lines that end with `.tgz` as packages when they aren't.
I suggest investigating a bit whether we can know how many packages we expect to pack, and fail if we find more than that.

EyalDelarea:
> This generally LGTM. The only problem with this approach is that we might unintentionally identify lines that end with `.tgz` as packages when they aren't. I suggest investigating a bit whether we can know how many packages we expect to pack, and fail if we find more than that.

So i've added a step to verify that each of the .tgz tarballs we identified from the output actually exists.

The only way i currently think of a .tgz suffix can enter the log is by a prepack script containing .tgz suffix for some reason...
that's why it will be ignored as the file doesn't really exists.

and if we have a .tgz file inside our folder that is not related, we won't publish it, as it wasn't created by the output of the pack command.

PR #1178, opened 2024-04-30
Update web login prompts

yahavi:
-----

Revise the web login commands errors (`jf login` & `jf c add` with Web Login) to advise executing `jf c add` if the Artifactory version is below 7.64.0. 
image

### DIFF
--- a/artifactory/utils/weblogin.go
+++ b/artifactory/utils/weblogin.go
@@ -28,8 +28,8 @@ func DoWebLogin(serverDetails *config.ServerDetails) (token auth.CommonTokenPara
 	}
 	if err = accessManager.SendLoginAuthenticationRequest(uuidStr); err != nil {
 		err = errors.Join(err,
-			errorutils.CheckErrorf("The 'Web Login' functionality is only supported for Artifactory version 7.64.0 and above. "+
-				"Make sure the details you entered are correct and that Artifactory meets the version requirement."))
+			errorutils.CheckErrorf("it looks like the Web Login is unavailable. This could be because your JFrog Artifactory version is less than 7.64.0. "+
+				"You can use the \"jf c add\" command to authenticate with the JFrog Platform"))
 		return
 	}
 	log.Info("After logging in via your web browser, please enter the code if prompted: " + uuidStr[len(uuidStr)-4:])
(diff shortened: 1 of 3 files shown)


eyalbe4 (changes requested):
Few comments to improve the user experience;

JFrog Platform URL:
-->
Enter your JFrog Platform URL:

Let's completely remove the question about the Artifactory Server proxy support. This is yet another adoption optical. Let's allow sending the proxy URL as a flag to the "jf c add" command only. Users who use the "jf login" shouldn't be bothered by it.

artifactory/utils/weblogin.go
```diff
@@ -28,8 +28,8 @@ func DoWebLogin(serverDetails *config.ServerDetails) (token auth.CommonTokenPara
 	}
 	if err = accessManager.SendLoginAuthenticationRequest(uuidStr); err != nil {
 		err = errors.Join(err,
-			errorutils.CheckErrorf("The 'Web Login' functionality is only supported for Artifactory version 7.64.0 and above. "+
-				"Make sure the details you entered are correct and that Artifactory meets the version requirement."))
+			errorutils.CheckErrorf("it looks like the Web Login is unavailable. This could be because your JFrog Artifactory version is less than 7.64.0. "+
+				"You can use the \"jf c add\" command to authenticate with the JFrog Platform"))
```
eyalbe4, line 32: errorutils.CheckErrorf("it looks like the Web Login is unavailable. This could be because your JFrog Artifactory version is less than 7.64.0. "+
				"You can use the \"jf c add\" command to authenticate with the JFrog Platform"))
-->
errorutils.CheckErrorf("oops! it looks like the web login option is unavailable right now. This could be because your JFrog Artifactory version is less than 7.64.0. \n"+
				"Don't worry! You can use the \"jf c add\" command to authenticate with the JFrog Platform instead"))

eyalbe4 (changes requested):

artifactory/utils/weblogin.go
yahavi: I changed it to:

errorutils.CheckErrorf("Oops! it looks like the web login option is unavailable right now. This could be because your JFrog Artifactory version is less than 7.64.0. \n"+
"Don't worry! You can use the "jf c add" command to authenticate with the JFrog Platform using other methods"))

eyalbe4: Looks good @yahavi.

PR #1187, opened 2024-05-23
Utils enhancements to allow go support for JFrog Curation

asafambar:
-----

Depends on: https://github.com/jfrog/build-info-go/pull/255


attiasas (commented):
* Please add test cases for the new bool argument of `getArtifactoryApiUrl`

artifactory/commands/golang/go.go
```diff
@@ -330,19 +330,25 @@ func buildPackageVersionRequest(name, branchName string) string {
...
-	repoUrl += "|direct"
+	if !isCurationCmd {
+		repoUrl += "|direct"
+	}
 	return os.Setenv("GOPROXY", repoUrl)
 }
+
+func SetGoModCache(cacheFolder string) error {
+	return os.Setenv("GOMODCACHE", cacheFolder)
+}
```
attiasas, line 354:
```suggestion
const GoModCacheEnvVar = "GOMODCACHE"
func SectGoModCache(cacheFolder string) (err error, restore func() error) {
	oldVal, exists := os.LookupEnv(GoModCacheEnvVar)
	return errorutils.CheckError(os.Setenv(GoModCacheEnvVar, cacheFolder)), func() error {
		if !exists {
			return errorutils.CheckError(os.Unsetenv(GoModCacheEnvVar))
		} else {
			return errorutils.CheckError(os.Setenv(GoModCacheEnvVar, oldVal))
		}
	}
}
```
You can add option to restore env when done, most of the times when we end our commands we want to restore the user env

asafambar: it shouldnt change the user env, only in this session maybe, but I'm not sure

utils/golang/utils.go
```diff
@@ -83,8 +84,8 @@ func GetModuleName(projectDir string) (string, error) {
 	return path, nil
 }
 
-func GetDependenciesList(projectDir string) (map[string]bool, error) {
-	deps, err := utils.GetDependenciesList(projectDir, log.Logger)
+func GetDependenciesList(projectDir string, errorFunc utils.HandleErrorFunc) (map[string]bool, error) {
```
eyalbe4, line 87: This function seems to be unused, so let's remove it.
The same goes for the below `GetDependenciesGraph` function. Let'e remove it as well.

utils/golang/utils.go
```diff
@@ -99,17 +100,41 @@ func GetDependenciesGraph(projectDir string) (map[string][]string, error) {
...
 		return "", err
 	}
-	return getArtifactoryApiUrl(repo, authServerDetails)
+	return getArtifactoryApiUrl(repo, authServerDetails, goProxyParams)
 }
 
 // Gets the URL of the specified repository Go API in Artifactory.
 // The URL contains credentials (username and access token or password).
-func getArtifactoryApiUrl(repoName string, details auth.ServiceDetails) (string, error) {
+func getArtifactoryApiUrl(repoName string, details auth.ServiceDetails, goProxyParams GoProxyUrlParams) (string, error) {
```
eyalbe4, line 137: You enhanced this function by adding a new `goProxyParams` argument to it, however, you did not enhance the existing `TestGetArtifactoryApiUrl` function to also test new argument functionality.

eyalbe4 (changes requested):
See my inline comments.
I'd love to run another review cycle following the implementation of my comments.

eyalbe4 (changes requested):
Let's also create a PR for github.com/jfrog/jfrog-cli with an upgrade of the jfrog-cli-core and build-info-go modules that include the changes proposed in their PRs, and ensure the tests there are passing.
Let's link that jfrog-cli PR to this PR and the build-info-go PR.

utils/golang/utils.go
asafambar: It's in use in the `security` repo

asafambar:
> Let's also create a PR for github.com/jfrog/jfrog-cli with an upgrade of the jfrog-cli-core and build-info-go modules that include the changes proposed in their PRs, and ensure the tests there are passing. Let's link that jfrog-cli PR to this PR and the build-info-go PR.

@attiasas  will take care of that

utils/golang/utils.go
eyalbe4: If those functions are used by the jfrog-cli-security module, let's move them to that module.

asafambar: Done

PR #1188, opened 2024-05-29
Support evidence service

lesnerd:
-----

### DIFF
--- a/artifactory/utils/utils.go
+++ b/artifactory/utils/utils.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	ioutils "github.com/jfrog/gofrog/io"
+	"github.com/jfrog/jfrog-client-go/evidence"
 	"io"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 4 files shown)


artifactory/utils/utils.go
```diff
@@ -214,6 +215,27 @@ func CreateLifecycleServiceManager(serviceDetails *config.ServerDetails, isDryRu
 	return lifecycle.New(serviceConfig)
 }
 
+func CreateEvidenceServiceManager(serviceDetails *config.ServerDetails, isDryRun bool) (*evidence.EvidenceServicesManager, error) {
```
eyalbe4, line 218: What is the reason this service isn't part of the jfrog-cli-artifactory Embedded Plugin?

eyalbe4 (changes requested):

artifactory/utils/utils.go
lesnerd: Since its a 3rd party api it made sense to put it to jfrog-client-go, hence the service manager had to be here.

lesnerd: More over the evidence service can not "live" logically without Artifactory, in that sense it is alike "ReleaseLifecycleManagement".

eyalbe4: Yeah, the service itself should be in jfrog-client-go, no argument there, but my question is why should the "CreateEvidenceServiceManager" function shouldn't be in jfrog-cli-artifactory? It is used only by the jfrog-cli-artifactory, so what's the reason for not placing it there?

eyalbe4: The fact that "ReleaseLifecycleManagement" isn't placed inside jfrog-cli-artifactory is because there's code in jfrog-cli-core that depends on it. The kdea is reduce the dependencies between modules to the minimum. 
Placing a single function in a module that doesn't use it doesn't make much sense.

RobiNino: After our face to face discussion we agreed we can currently merge this code to the core due to the effort required by adding it somewhere else, and because the core is a shared component.

RobiNino (approved):

PR #1189, opened 2024-06-02
Move techutils to the jfrog-cli-security module

attiasas:
-----

Depends on: https://github.com/jfrog/jfrog-cli-core/pull/1190
Moving this content to the jfrog-security-cli project


eyalbe4 (changes requested):
The technologies code should be moved to the jfrog-cli-security module. 
The more we wait with this move, the harder it'll be to implement in the future.

attiasas:
@eyalbe4, I can't move this since you are also using it here:
https://github.com/jfrog/jfrog-cli-core/blob/master/general/project/projectinit.go#L52

utils/coreutils/techutils.go
```diff
@@ -88,51 +105,51 @@ var technologiesData = map[Technology]TechData{
...
 		packageDescriptors:         []string{"go.mod"},
 		packageVersionOperator:     "@v",
 		packageInstallationCommand: "get",
 	},
 	Pip: {
 		packageType:            Pypi,
-		indicators:             []string{"setup.py", "requirements.txt"},
-		packageDescriptors:     []string{"setup.py", "requirements.txt"},
-		exclude:                []string{"Pipfile", "Pipfile.lock", "pyproject.toml", "poetry.lock"},
+		indicators:             map[string]ContentValidator{"pyproject.toml": pyProjectTomlIndicatorContent(Pip), "setup.py": nil, "requirements.txt": nil},
```
hadarshjfrog, line 130: I think the current implementation is good - but IMO adding a key that'll indicate it - will be a bit more cleaner than having keys referring to nil values, most of the team.
The new key as it could be empty other than python packages - and only if relevant it'll be checked.
wdyt?

utils/coreutils/techutils_test.go
```diff
@@ -364,14 +374,46 @@ func TestCleanSubDirectories(t *testing.T) {
 	}
 }
 
+func createTempDirWithPyProjectToml(t *testing.T, tech Technology) (tmpDir string, callback func()) {
```
hadarshjfrog, line 377: can we add a test for a project with multiple descriptors (specifically with PyProjectToml) but that one of them isn't python? e.g - requirements.txt, package.json and PyProject.toml - could be interesting to validate...

PR #1201, opened 2024-07-15
Fix root module name handling in build info collection

EyalDelarea:
-----

## Bug Description:
When collecting build information for different packages, if the module name is not specified in the descriptor file or provided as a CLI argument, it defaults to an empty string. This results in an error from the Artifactory server, stating that the module ID cannot be null.

```
16:22:54 [Error] server response: 400 
{
  "errors": [
    {
      "status": 400,
      "message": "Module id name cannot be null!"
    }
  ]
}
```

## Fix:
If no module ID is provided, the default will now be set to the base directory of the module. This aligns with the default behavior of package descriptors init scripts.
As a last reasort the module name will resolve to "module" to be displayed at the UI and not fail on publish.


artifactory/commands/npm/npmcommand.go
```diff
@@ -402,6 +401,15 @@ func (nc *NpmCommand) GetRepo() string {
 	return nc.repo
 }
 
+// Resolves to the base directory name if a module name is not provided in package.json or CLI command.
```
eyalbe4, line 404: Resolves to the base directory name ... --> Returns the project's root directory name ...

Let's also add to function's description the case where the module name is provided.

eyalbe4 (changes requested):

artifactory/commands/npm/npmcommand.go
```diff
@@ -402,6 +401,15 @@ func (nc *NpmCommand) GetRepo() string {
...
+func (nc *NpmCommand) resolveModuleName() string {
+	moduleName := nc.buildConfiguration.GetModule()
+	if moduleName == "" {
+		moduleName = path.Base(nc.workingDirectory)
```
yahavi, line 408: Shouldn't be filepath.Join?

EyalDelarea:
@yahavi i've changed the function to return error if fails to get working dir.
let me know what you think 🚀

common/build/buildutils.go
```diff
@@ -461,6 +462,22 @@ func (bc *BuildConfiguration) IsLoadedFromConfigFile() bool {
...
+// If the working directory cannot be retrieved, it defaults to "module".
+// Module name is mandatory to publish build info.
+func (bc *BuildConfiguration) ResolveBaseModuleName() (moduleName string, err error) {
+	if bc.module != "" {
+		moduleName = bc.module
+		return
+	}
+	wd, err := os.Getwd()
+	if err != nil {
+		return
```
yahavi, line 476: Let's put here checkerr and remove it from all other users of ResolveBaseModuleName

kunkka-m:
image
image
I used the latest version of jfrog-cli(2.63.0), but this issue still happened

kunkka-m:
Oh, This MR is to dev branch!

lincoohe:
Hi @EyalDelarea ,it should be the **build-publish** issue  instead of **bce** issue.

PR #1217, opened 2024-07-30
embedded plugin context - don't set flags if not exists

attiasas:
-----

For embedded plugins, when using its `context` flags where reported as set when they are not.
Fixing: `(c *Context) IsFlagSet(flagName string) bool`

### DIFF
--- a/plugins/components/conversionlayer.go
+++ b/plugins/components/conversionlayer.go
@@ -426,6 +426,7 @@ func getValueForStringFlag(f StringFlag, baseContext *cli.Context) (finalValue s
 		return
 	}
 	if f.DefaultValue != "" {
+		// Empty but has a default value defined.
 		finalValue = f.DefaultValue
 		return
 	}
--- a/plugins/components/conversionlayer.go
+++ b/plugins/components/conversionlayer.go
@@ -408,11 +408,9 @@ func fillFlagMaps(c *Context, baseContext *cli.Context, originalFlags []Flag) er
 		}
 
 		if boolFlag, ok := flag.(BoolFlag); ok {
-			if baseContext.IsSet(boolFlag.Name) {
-				finalValue, skip := getValueForBoolFlag(boolFlag, baseContext)
-				if !skip {
-					c.boolFlags[boolFlag.Name] = finalValue
-				}
+			finalValue, skip := getValueForBoolFlag(boolFlag, baseContext)
+			if !skip {
+				c.boolFlags[boolFlag.Name] = finalValue
 			}
 		}
 	}
--- a/plugins/components/conversionlayer.go
+++ b/plugins/components/conversionlayer.go
@@ -397,6 +397,10 @@ func fillFlagMaps(c *Context, baseContext *cli.Context, originalFlags []Flag) er
 	// Loop over all plugin's known flags.
 	for _, flag := range originalFlags {
 		if stringFlag, ok := flag.(StringFlag); ok {
+			if !baseContext.IsSet(stringFlag.Name) {
+				// Flag not set, continue.
+				continue
+			}
 			finalValue, err := getValueForStringFlag(stringFlag, baseContext.String(stringFlag.Name))
 			if err != nil {
 				return err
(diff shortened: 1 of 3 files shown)


RobiNino (approved):

plugins/components/conversionlayer.go
```diff
@@ -397,39 +397,53 @@ func fillFlagMaps(c *Context, baseContext *cli.Context, originalFlags []Flag) er
...
 
-func getValueForBoolFlag(f BoolFlag, baseContext *cli.Context) bool {
+func getValueForBoolFlag(f BoolFlag, baseContext *cli.Context) (finalValue, skip bool) {
 	if f.DefaultValue {
-		return baseContext.BoolT(f.Name)
+		finalValue = baseContext.BoolT(f.Name)
+		return
 	}
-	return baseContext.Bool(f.Name)
+	skip = !baseContext.IsSet(f.Name)
```
RobiNino, line 446: I'm not sure we want to skip adding a bool flag with default value false, this is a legitimate value. Maybe we shouldn't skip bool flags at all.

plugins/components/conversionlayer.go
```diff
@@ -397,39 +397,53 @@ func fillFlagMaps(c *Context, baseContext *cli.Context, originalFlags []Flag) er
...
+			if skip {
+				continue
+			}
 			c.stringFlags[stringFlag.Name] = finalValue
-			continue
 		}
 
 		if boolFlag, ok := flag.(BoolFlag); ok {
-			c.boolFlags[boolFlag.Name] = getValueForBoolFlag(boolFlag, baseContext)
+			if baseContext.IsSet(boolFlag.Name) {
```
RobiNino, line 411: We probably want to always add value for bool flags, not only when they are set.

RobiNino (changes requested):

plugins/components/conversionlayer.go
```diff
@@ -397,34 +397,37 @@ func fillFlagMaps(c *Context, baseContext *cli.Context, originalFlags []Flag) er
...
 }
 
-func getValueForStringFlag(f StringFlag, receivedValue string) (finalValue string, err error) {
-	if receivedValue != "" {
-		return receivedValue, nil
+func getValueForStringFlag(f StringFlag, baseContext *cli.Context) (finalValue string, err error) {
+	value := baseContext.String(f.Name)
+	if value != "" {
+		finalValue = value
+		return
```
RobiNino, line 419:
```suggestion
		return value
```

plugins/components/conversionlayer.go
```diff
@@ -397,34 +397,37 @@ func fillFlagMaps(c *Context, baseContext *cli.Context, originalFlags []Flag) er
...
 			c.boolFlags[boolFlag.Name] = getValueForBoolFlag(boolFlag, baseContext)
 		}
 	}
 	return nil
 }
 
-func getValueForStringFlag(f StringFlag, receivedValue string) (finalValue string, err error) {
-	if receivedValue != "" {
-		return receivedValue, nil
+func getValueForStringFlag(f StringFlag, baseContext *cli.Context) (finalValue string, err error) {
```
RobiNino, line 415:
```suggestion
func getValueForStringFlag(f StringFlag, baseContext *cli.Context) (string, error) {
```

plugins/components/conversionlayer.go
```diff
@@ -397,34 +397,37 @@ func fillFlagMaps(c *Context, baseContext *cli.Context, originalFlags []Flag) er
...
+	if value != "" {
+		finalValue = value
+		return
 	}
-	// Empty but has a default value defined.
 	if f.DefaultValue != "" {
-		return f.DefaultValue, nil
+		// Empty but has a default value defined.
+		finalValue = f.DefaultValue
+		return
```
RobiNino, line 424:
```suggestion
		return f.DefaultValue
```

plugins/components/conversionlayer.go
```diff
@@ -397,34 +397,37 @@ func fillFlagMaps(c *Context, baseContext *cli.Context, originalFlags []Flag) er
...
 	}
-	// Empty but mandatory.
 	if f.Mandatory {
-		return "", errors.New("Mandatory flag '" + f.Name + "' is missing")
+		// Empty but mandatory.
+		err = errors.New("Mandatory flag '" + f.Name + "' is missing")
```
RobiNino, line 428:
```suggestion
		return "", errors.New("Mandatory flag '" + f.Name + "' is missing")
```

plugins/components/conversionlayer.go
```diff
@@ -397,34 +397,37 @@ func fillFlagMaps(c *Context, baseContext *cli.Context, originalFlags []Flag) er
...
 	}
-	return "", nil
+	return
```
RobiNino, line 430:
```suggestion
	return "", nil
```

RobiNino (changes requested):

PR #1225, opened 2024-08-07
Recognize "403 forbidden" error from pip command output, and clasiffy it as "forbidden" error type

asafambar:
-----
Description:
After running "pip <command>" and encountering an error, we examine the output from stderr. If the output includes a "403 Forbidden", we wrap the error with a specific error type for "Forbidden 403." This error type will be recognized later by post actions, such as curation.


artifactory/commands/python/python.go
```diff
@@ -71,7 +72,24 @@ func (pc *PythonCommand) Run() (err error) {
...
+		multiWriter := io.MultiWriter(os.Stderr, errBuffer)
+		cmd.Stderr = multiWriter
+		cmd.Stdout = os.Stdout
+
+		err = cmd.Run()
+		if err != nil {
+			if buildInfoUtils.IsForbiddenOutput("pip", errBuffer.String()) {
+				err = errors.Join(err, buildInfoUtils.NewForbiddenError())
+			}
+		}
```
eyalbe4, line 92: The above code should be mioved into in the build-info-go module. If it isn't place there, this functionality will not work for Pythin when collecting build-info.
It is true that when not collecting build-info for python, the code flow ends here and doesn't continue into build-info-go, but having this code there, will allow you to invoke it from here as well.

eyalbe4 (changes requested):

artifactory/commands/python/python.go
asafambar: Eyal, In case we collect build info, the whole installation have different logic which apparently adding the stderr to the err, so in this case it will work in the flow of pip install + collecting dependencies for build-info, I security wrapper we recognize either a 403 in the error string OR forbidden type error.

I don't want to use the same logic to run installation command for the build info, as it can break existing behavior, which I avoided in this PRs.

PR #1226, opened 2024-08-11
Mvn - avoid parsing build info when it has not been collected

omerzi:
-----

### DIFF
--- a/artifactory/commands/mvn/mvn.go
+++ b/artifactory/commands/mvn/mvn.go
@@ -265,9 +265,10 @@ func (mc *MvnCommand) updateBuildInfoArtifactsWithDeploymentRepo(vConfig *viper.
 	if err != nil {
 		return errorutils.CheckErrorf("failed to read build info file: %s", err.Error())
 	}
-	// There are some scenarios where the build info details were set but no content was generated,
-	// such as running a `jf mvn <command>` from the jfrog-cli-setup GitHub Action,
-	// which automatically fills the build info details, but the command itself doesn't always generate any content.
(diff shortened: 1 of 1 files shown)


artifactory/commands/mvn/mvn.go
```diff
@@ -153,9 +153,15 @@ func (mc *MvnCommand) Run() error {
 		return err
 	}
 
-	if err = mc.updateBuildInfoArtifactsWithDeploymentRepo(vConfig, mvnParams.GetBuildInfoFilePath()); err != nil {
+	isCollectedBuildInfo, err := mc.configuration.IsCollectBuildInfo()
```
yahavi, line 156: Is `mc.configuration` always not null?

artifactory/commands/mvn/mvn.go
```diff
@@ -257,6 +263,9 @@ func (mc *MvnCommand) updateBuildInfoArtifactsWithDeploymentRepo(vConfig *viper.
 	if err != nil {
 		return errorutils.CheckErrorf("failed to read build info file: %s", err.Error())
 	}
+	if len(content) == 0 {
```
sverdlov93, line 266: when content can be 0?
Is it a possible use case or an error we should throw?

artifactory/commands/mvn/mvn.go
```diff
@@ -257,6 +263,9 @@ func (mc *MvnCommand) updateBuildInfoArtifactsWithDeploymentRepo(vConfig *viper.
...
+		return nil
```
yahavi, line 267: Let's add a comment explaining when it can be empty

PR #1230, opened 2024-08-14
Command Summary - New record functionality

EyalDelarea:
-----

In addition to the existing simple data recording function, we now have the capability to record data with a predefined 'index'. This allows us to categorize and store specific information that can be easily accessed by other components within the summary report.

The main example of this usage is storing indexed scan results, which can later be attached to specific files in the summary.

Here’s an example of how to use the new function:

```go
err = buildInfoSummary.RecordWithIndex("myBuildScanData", BuildScan, BuildName,BuildNumber)
```

Will resolve to a file system like this:
image

The supported indexes so far are:
1. Build Scan
2. Binaries Scan
3. Docker Scan
4. Sarif Report

### Note

This change also requires that the generation of markdown will happen via a CLI command, rather than automatically. The PR in the CLI will introduce a new command called "create-summary-markdown".


utils/config/config.go
```diff
@@ -41,7 +41,7 @@ func IsServerConfExists() (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	return conf.Servers != nil && len(conf.Servers) > 0, nil
+	return len(conf.Servers) > 0, nil
```
EyalDelarea, line 44: This is due golint change

eyalbe4 (changes requested):
Shouldn't there be a new JFrog CLI command that invokes the new `GenerateMarkdown` API?

EyalDelarea:
> Shouldn't there be a new JFrog CLI command that invokes the new `GenerateMarkdown` API?

Yes, it is presented here.
https://github.com/jfrog/jfrog-cli/pull/2653

commandsummary/commandsummary.go
```diff
@@ -8,19 +8,39 @@ import (
 	"github.com/jfrog/jfrog-client-go/utils/errorutils"
 	"github.com/jfrog/jfrog-client-go/utils/io/fileutils"
 	"os"
-	"path"
 	"path/filepath"
+	"regexp"
+	"runtime"
 	"strings"
 )
```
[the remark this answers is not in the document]
EyalDelarea: Done

commandsummary/commandsummary.go
```diff
@@ -144,6 +201,28 @@ func ShouldRecordSummary() bool {
 	return os.Getenv(coreutils.OutputDirPathEnv) != ""
 }
 
+func createAndWriteToFile(filePath, fileName string, data []byte) (err error) {
```
eyalbe4, line 204: Let's add a description for this function and its arguments.
In addition, let's add comments that explain its code flow, because the logic of the below code flow isn;t quite clear - 
```
	if strings.Contains(fileName, "*") {
		fd, err = os.CreateTemp(filePath, fileName)
	} else {
		fd, err = os.Create(filepath.Join(filePath, fileName))
	}
```

eyalbe4 (changes requested):

PR #1235, opened 2024-08-14
Replace poetry update with poetry install

BartSchuurmans:
`poetry update` is a command that upgrades all dependecies to their latest compatible version. This is never what you want when just installing dependencies. Instead, `poetry install` should be run to install the dependency versions specified in the lock file.

This fixes the behavior of Frogbot upgrading all dependencies when called in CI.

-----


BartSchuurmans:

BartSchuurmans:
recheck

BartSchuurmans:
Could someone take a look at this? My company is a paying customer of JFrog but has difficulties using it because we widely use poetry and the current behaviour is not correct.

PR #1261, opened 2024-09-05
Allow setting margin to table frame

sverdlov93:
-----


utils/coreutils/tableutils.go
```diff
@@ -370,12 +370,18 @@ type embeddedTableCell struct {
 // ┌─────────────────────────────────────────┐
 // │ An example of a message in a nice frame │
 // └─────────────────────────────────────────┘
-func PrintMessage(message string) {
+func PrintMessageInsideFrame(message, marginLeft string) {
```
omerzi, line 373: Let's add a comment above this function with an example of how the `marginLeft` value can change the frame view.

PR #1267, opened 2024-09-18
Add twine support

RobiNino:
-----
Depends on https://github.com/jfrog/build-info-go/pull/276

### DIFF
--- a/artifactory/commands/python/twine.go
+++ b/artifactory/commands/python/twine.go
@@ -170,8 +170,8 @@ func (tc *TwineCommand) uploadAndCollectBuildInfo() error {
 	if err != nil {
 		return err
 	}
-	for _, artifact := range artifacts {
-		artifact.OriginalDeploymentRepo = tc.targetRepo
+	for i := range artifacts {
+		artifacts[i].OriginalDeploymentRepo = tc.targetRepo
 	}
 	if err = pythonModule.AddArtifacts(artifacts); err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


artifactory/commands/python/twine.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+func (tc *TwineCommand) runPlainTwineCommand() error {
+	log.Debug("Running twine command:", tc.commandName, strings.Join(tc.args, " "))
+	args := append([]string{tc.commandName}, tc.args...)
+	cmd := exec.Command(_twineExecName, args...)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	err := cmd.Run()
+	return err
```
eyalbe4, line 145: Can be simplified to - 
```
return cmd.Run()
```

artifactory/commands/python/twine.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+func (tc *TwineCommand) uploadAndCollectBuildInfo() error {
+	buildInfo, err := buildUtils.PrepareBuildPrerequisites(tc.buildConfiguration)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		if buildInfo != nil && err != nil {
```
eyalbe4, line 155: What's the reason we're cheching here whether err isn't nil?
Isn't it enough to check that buildInfo isn't nil?

eyalbe4 (changes requested):

PR #1287, opened 2024-10-30
Clean code, remove redundant, and unite common build tools logic

sverdlov93:
-----

**No new code here, only movements.**


artifactory/commands/python/python.go
```diff
@@ -154,3 +169,111 @@ func (pc *PythonCommand) GetStdWriter() io.WriteCloser {
 func (pc *PythonCommand) GetErrWriter() io.WriteCloser {
 	return nil
 }
+
+func GetPypiRepoUrlWithCredentials(serverDetails *config.ServerDetails, repository string, isCurationCmd bool) (*url.URL, string, string, error) {
```
eyalbe4, line 173: Let's check if this function can be made private.
The same goes for the other new functions in this file.

eyalbe4 (changes requested):

PR #1299, opened 2024-11-10
Add commands to usage report

EyalDelarea:
-----
## This PR adds tracking for new command executions in the usage report:

`config_oidc`: Tracks the number of times the config command is executed with OIDC.
`config` : Track the amount of times JFrog server has been configured. ( Notice this includes Add or Edit operation for both OIDC and non OIDC )
`rt_build_publish_auto`: Tracks instances when builds are automatically published.

Tracking is enabled by setting specific environment variables in the `setup-cli-action`, which the CLI then reads to report usage data.

**Note**: The config command usage report only includes cases where `config_oidc` is used, as it is more useful to differentiate between "add" and "edit" operations rather than general config usage.


artifactory/commands/buildinfo/publish.go
```diff
@@ -70,6 +70,10 @@ func (bpc *BuildPublishCommand) IsDetailedSummary() bool {
 }
 
 func (bpc *BuildPublishCommand) CommandName() string {
+	isAutoPublishedTriggered, err := clientutils.GetBoolEnvValue(coreutils.UsageAutoPublishedBuild, false)
+	if err == nil && isAutoPublishedTriggered {
```
eyalbe4, line 74: If the error isn't nil, let's log the error, with a prefix that mention the context of the error.
The same applies for the `CommandName` function in the `config.go` file.

artifactory/commands/buildinfo/publish.go
```diff
@@ -70,6 +70,10 @@ func (bpc *BuildPublishCommand) IsDetailedSummary() bool {
...
+		return "rt_auto_build_publish"
```
eyalbe4, line 75: rt_auto_build_publish --> rt_build_publish_auto

common/commands/config.go
```diff
@@ -41,6 +41,13 @@ const (
 	WebLogin    AuthenticationMethod = "Web Login"
 )
 
+const (
+	// Indicates that the config command uses OIDC authentication.
+	ConfigOidcCommandName = "config_OIDC"
+	// Default config command name.
+	ConfigCommandName = "config"
```
eyalbe4, line 48: config_OIDC --> config_oidc
Let's check if the above two consts can be made private.

common/commands/config.go
```diff
@@ -143,7 +150,26 @@ func (cc *ConfigCommand) ServerDetails() (*config.ServerDetails, error) {
...
+	if err == nil && isOidcConfigured {
+		return ConfigOidcCommandName
+	}
+	return ConfigCommandName
+}
+
+// ExecAndReportUsage runs the ConfigCommand and then triggers a usage report if needed,
+// Report usage only if OIDC integration was used
+// Usage must be sent after command execution as we need the server details to be set.
+func (cc *ConfigCommand) ExecAndReportUsage() error {
```
eyalbe4, line 163: To simply things, I suggest we always report the usage here for the `jf config` command, regardless if OIDC is used or not. Of curse, we're using two different command names for OIDC and non-OIDC config.

common/commands/config.go
```diff
@@ -143,7 +150,26 @@ func (cc *ConfigCommand) ServerDetails() (*config.ServerDetails, error) {
...
+	err := cc.Run()
+	if cc.CommandName() == ConfigOidcCommandName {
+		channel := make(chan bool)
+		// Triggers the report usage.
+		go reportUsage(cc, channel)
```
eyalbe4, line 168: In this specific case, there's no need to run the `reportUsage` as a go routine, beucase there's nothing to parallelize here. This function's execution is triggered after the  `c.Run()` command anyway. There;s n need for the channel as well.
We can simply have -
```
if err := cc.Run(); err != nil {
   return err
}
reportUsage(cc, nil)
return nil
```
and then we can change the `reportUsage` function to handle a nil channel.

common/commands/config_test.go
```diff
@@ -344,6 +344,67 @@ func TestImport(t *testing.T) {
...
+func TestCommandName(t *testing.T) {
+	cc := NewConfigCommand(AddOrEdit, testServerId)
+
+	// Test when the environment variable is not set
+	err := os.Unsetenv(coreutils.UsageOidcConfigured)
+	assert.NoError(t, err)
+	assert.Equal(t, ConfigCommandName, cc.CommandName())
+
+	// Test when the environment variable is set
+	err = os.Setenv(coreutils.UsageOidcConfigured, "true")
```
eyalbe4, line 356: Let's use the `testsutils.SetEnvWithCallbackAndAssert`.
See it's usage in the code as a reference.

common/commands/config_test.go
```diff
@@ -344,6 +344,67 @@ func TestImport(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	assert.Equal(t, ConfigOidcCommandName, cc.CommandName())
+
+	// Clean up
+	err = os.Unsetenv(coreutils.UsageOidcConfigured)
+	assert.NoError(t, err)
+}
+
+func TestConfigCommand_ExecAndReportUsage(t *testing.T) {
```
eyalbe4, line 366: Let's add a defer function at the beginning of this function, that unsets the `UsageOidcConfigured` variable.

utils/coreutils/coreconsts.go
```diff
@@ -49,6 +49,9 @@ const (
 	CI                      = "CI"
 	ServerID                = "JFROG_CLI_SERVER_ID"
 	TransitiveDownload      = "JFROG_CLI_TRANSITIVE_DOWNLOAD"
+	// Envs used to determine command types for usage reports
```
eyalbe4, line 52: Let's add more information to this comment, explaining that those variables can be set by the `setup-jfrog-cli` GitHub Action, and also provide some background about the use case.
This will help future readers of this code understand the context.

eyalbe4 (changes requested):

PR #1314, opened 2024-12-20
Internal API - Visibility System support

eyalbe4:
This pull request introduces enhancements to the usage reporting mechanism in the JFrog CLI by adding a new VisibilitySystemManager to log metrics and integrating it into existing commands

Depends on https://github.com/jfrog/jfrog-client-go/pull/1060


common/commands/command.go
```diff
@@ -30,29 +33,59 @@ func Exec(command Command) error {
 }
 
 func reportUsage(command Command, channel chan<- bool) {
+	// When the usage reporting is donw, signal to the channel.
```
sverdlov93, line 36:
```suggestion
	// When the usage reporting is done, signal to the channel.
```

common/commands/command.go
```diff
@@ -30,29 +33,59 @@ func Exec(command Command) error {
...
+	serverDetails, err := command.ServerDetails()
+	if err != nil {
+		log.Debug("Usage reporting:", err.Error())
+		return
+	}
+
+	if serverDetails != nil {
+		var wg sync.WaitGroup
+
+		// Repotr the usage to Artifactory's Call Home API.
```
sverdlov93, line 53:
```suggestion
		// Report the usage to Artifactory's Call Home API.
```

common/commands/command.go
```diff
@@ -30,29 +33,59 @@ func Exec(command Command) error {
...
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				reportUsageToArtifactoryCallHome(command, serverDetails)
+			}()
 		}
-	} else {
-		log.Debug("Usage info is disabled.")
+
+		// Repotr the usage to the Visibility System.
```
sverdlov93, line 62:
```suggestion
		// Report the usage to the Visibility System.
```

utils/usage/visibility_system_manager.go
```diff
@@ -0,0 +1,68 @@
...
+	RunID                                string `json:"run_id"`
+	GitRepo                              string `json:"git_repo"`
+	GhTokenForCodeScanningAlertsProvided string `json:"gh_token_for_code_scanning_alerts_provided"`
+}
+
+type visibilityMetric struct {
+	Value       int    `json:"value"`
+	MetricsName string `json:"metrics_name"`
+	Labels      labels `json:"labels"`
+}
```
sverdlov93, line 36: **without** `omitempty` it will send here `Value 0` for empty value for example. If its a possible value a believe that its not good. same goes with empty strings.

**with** omitempty , it won't but it won't send actual 0 value if needed.

so it's need to be decided according to context.

utils/usage/visibility_system_manager.go
```diff
@@ -0,0 +1,68 @@
...
+		OIDCUsed:                             os.Getenv("JFROG_CLI_USAGE_OIDC_USED"),
+		JobID:                                os.Getenv("JFROG_CLI_USAGE_JOB_ID"),
+		RunID:                                os.Getenv("JFROG_CLI_USAGE_RUN_ID"),
+		GitRepo:                              os.Getenv("JFROG_CLI_USAGE_GIT_REPO"),
+		GhTokenForCodeScanningAlertsProvided: os.Getenv("JFROG_CLI_USAGE_GH_TOKEN_FOR_CODE_SCANNING_ALERTS_PROVIDED"),
+	}
+
+	metric := visibilityMetric{
+		Value:       1,
+		MetricsName: "jfcli_commands_countaa",
```
sverdlov93, line 51: just checking that _countaa is not a typo

utils/usage/visibility_system_manager.go
```diff
@@ -0,0 +1,68 @@
...
+func (vsm *VisibilitySystemManager) SendUsage(commandName string) error {
+	manager, err := utils.CreateJfConnectServiceManager(vsm.serverDetails)
+	if err != nil {
+		return err
+	}
+	metric, err := vsm.createMetric(commandName)
+	if err != nil {
+		return err
+	}
+	return manager.LogMetric(metric)
```
sverdlov93, line 67: On jfrog cli context LogMetric may sound and some local Log func. LogDebug or something like that.

utils/usage/visibility_system_manager.go
```diff
@@ -0,0 +1,68 @@
...
+	GhTokenForCodeScanningAlertsProvided string `json:"gh_token_for_code_scanning_alerts_provided"`
+}
+
+type visibilityMetric struct {
+	Value       int    `json:"value"`
+	MetricsName string `json:"metrics_name"`
+	Labels      labels `json:"labels"`
+}
+
+func (vsm *VisibilitySystemManager) createMetric(commandName string) ([]byte, error) {
```
sverdlov93, line 38: why dont we use VisibilityMetric as a return type, to enforce this type on the manager.LogMetric() func?

utils/usage/visibility_system_manager.go
eyalbe4: Wow. Great catch @sverdlov93!!

utils/usage/visibility_system_manager.go
eyalbe4: Thanks @sverdlov93.
Omit empty won't be good in the case for the following reasons:
1. We'd like to send empty strings in the JSON payload, if no  values are sent. I believe that the payload structure is expected to be consistent.
2. The value that we sent is always 1, so it'll never be empty.

utils/usage/visibility_system_manager.go
eyalbe4: That indeed makes sense @sverdlov93. Thanks!
Let me implement this in a follow-up PR, after this upcoming initial release.

utils/config/config.go
```diff
@@ -747,6 +748,12 @@ func (serverDetails *ServerDetails) CreateAccessAuthConfig() (auth.ServiceDetail
 	return serverDetails.createAuthConfig(pAuth)
 }
 
+func (serverDetails *ServerDetails) CreateJfConnectAuthConfig() (auth.ServiceDetails, error) {
+	pAuth := accessAuth.NewAccessDetails()
+	pAuth.SetUrl(utils.AddTrailingSlashIfNeeded(serverDetails.Url) + "jfconnect/")
```
sverdlov93, line 753: woudln't this cause double `jfconnect/` on the URL?
because here you create the base URL `ecosys.jfrog.io/access/jfconnect` and on the PostMetric you add `jfconnect/api/v1/backoffice/metrics/log`

eyalbe4: I think you're right @sverdlov93. Good catch 👍

PR #1319, opened 2024-12-24
Internal - Modify the Markdown Summary Artifactory links

eyalbe4:
Modify the links of the command's Markdown Summary, so that they include additional query params, that will be captured by the JFrog Platform backend.


artifactory/utils/commandsummary/buildinfosummary_test.go
```diff
@@ -1,15 +1,16 @@
...
-	"github.com/jfrog/jfrog-cli-core/v2/utils/coreutils"
-	"github.com/stretchr/testify/assert"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
 	"testing"
+
+	buildInfo "github.com/jfrog/build-info-go/entities"
+	buildinfo "github.com/jfrog/build-info-go/entities"
```
yahavi, line 11: package "github.com/jfrog/build-info-go/entities" is being imported more than once

artifactory/utils/commandsummary/buildinfosummary.go
```diff
@@ -2,13 +2,15 @@ package commandsummary
 
 import (
 	"fmt"
+	"net/url"
+	"path"
+	"strings"
+
+	"github.com/jfrog/build-info-go/entities"
```
yahavi, line 9: package "github.com/jfrog/build-info-go/entities" is being imported more than once

artifactory/utils/commandsummary/utils.go
```diff
@@ -33,3 +40,29 @@ func fileNameToSha1(fileName string) string {
 	hashBytes := hash.Sum(nil)
 	return hex.EncodeToString(hashBytes)
 }
+
+// addGitHubTrackingToUrl adds GitHub-related query parameters to a given URL if the GITHUB_WORKFLOW environment variable is set.
+func addGitHubTrackingToUrl(urlStr, section string) (string, error) {
```
yahavi, line 45: It seems that the section can only be either 'artifacts' or 'packages'. Using an enum would make it safer to manage this case.
```suggestion
func addGitHubTrackingToUrl(urlStr string, section summarySection) (string, error) {
```

```golang
type summarySection string

const (
	artifactsSection summarySection = "artifacts"
	packagesSection                 = "packages"
)
```

yahavi (changes requested):

PR #1324, opened 2024-12-31
Test Cli outputs of cmd with errors

attiasas:
-----

Add option to run test CLI with command that return error as well

### DIFF
--- a/utils/tests/test_cli.go
+++ b/utils/tests/test_cli.go
@@ -52,21 +52,32 @@ func (cli *JfrogCli) RunCliCmdWithOutput(t *testing.T, args ...string) string {
 	return RunCmdWithOutput(t, func() error { return cli.Exec(args...) })
 }
 
-// Run a command, redirect the stdout and return the output
-func RunCmdWithOutput(t *testing.T, executeCmd func() error) string {
+func (cli *JfrogCli) RunCliCmdWithOutputs(t *testing.T, args ...string) (string, error) {
+	return RunCmdWithOutputs(t, func() error { return cli.Exec(args...) })
+}
+
+func RunCmdWithOutputs(t *testing.T, executeCmd func() error) (output string, err error) {
 	newStdout, stdWriter, cleanUp := redirectOutToPipe(t)
 	defer cleanUp()
 
 	go func() {
-		assert.NoError(t, executeCmd())
+		err = executeCmd()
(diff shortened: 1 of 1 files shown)


utils/tests/test_cli.go
```diff
@@ -52,21 +52,32 @@ func (cli *JfrogCli) RunCliCmdWithOutput(t *testing.T, args ...string) string {
...
+	return RunCmdWithOutputs(t, func() error { return cli.Exec(args...) })
+}
+
+func RunCmdWithOutputs(t *testing.T, executeCmd func() error) (output string, err error) {
 	newStdout, stdWriter, cleanUp := redirectOutToPipe(t)
 	defer cleanUp()
 
 	go func() {
-		assert.NoError(t, executeCmd())
+		err = executeCmd()
```
sverdlov93, line 64: it a gorutine. so this error won't return

utils/tests/test_cli.go
```diff
@@ -52,21 +52,32 @@ func (cli *JfrogCli) RunCliCmdWithOutput(t *testing.T, args ...string) string {
...
 		// Closing the temp stdout in order to be able to read it's content.
 		assert.NoError(t, stdWriter.Close())
 	}()
 
-	content, err := io.ReadAll(newStdout)
+	content, e := io.ReadAll(newStdout)
+	assert.NoError(t, e)
+	output = string(content)
+	log.Debug(output)
```
sverdlov93, line 72: this will make the test logs in github to long to read

attiasas: this logic is not changed, stayed as it was before, should I remove this?

utils/tests/test_cli.go
attiasas: its a named output variable, also we are using the output and reading all the content right after running it.
maybe `io.ReadAll` is blocking? I did not change the logic just extracted it to a method

utils/tests/test_cli.go
sverdlov93: you are right. i remember changing this to debug for that reason

utils/tests/test_cli.go
attiasas: used channel to pass the error to make sure

sverdlov93 (approved):

PR #1334, opened 2025-01-17
Migrate artifactory commands to jfrog-cli-artifactory

bhanurp:
-----

This pull request depends on the following PR:
- [jfrog-cli-artifactory#34](https://github.com/jfrog/jfrog-cli-artifactory/pull/34)


go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/jedib0t/go-pretty/v6 v6.6.5
 	github.com/jfrog/build-info-go v1.10.8
 	github.com/jfrog/gofrog v1.7.6
+	github.com/jfrog/jfrog-cli-artifactory v0.1.11
```
RobiNino, line 17: Let's try to avoid this cyclic import.

RobiNino (changes requested):
Nice work!
Leaving most of my review as a comment since it involves unmodified files.

1. artifactory/commands/gradle - should be deleted.
2. artifactory/commands/testdata - should be deleted.
3. Transfer command - I think those should move too, because they belong in the Artifactory repo too, even if they will fall under another team's management. This will also help avoiding the cyclic import which is problematic. Should be moved with the relevant utils in artifactory/commands/utils (precheck, and the transfer named ones)
4. The tests directory that was migrated should also be deleted.

RobiNino:
It provides as a wrapping to all unit tests, so we can set a temporary jfrog home directory and set a logger. Let's currently leave this file as long as there are unit tests in this repository.
Look at it again, the tests workflow might still be required, maybe just modified.

go.mod
```diff
@@ -89,15 +88,18 @@ require (
...
 )
 
-// replace github.com/jfrog/jfrog-client-go => github.com/jfrog/jfrog-client-go v1.28.1-0.20250121093504-a3e981c875d7
+// replace github.com/jfrog/jfrog-client-go => github.com/jfrog/jfrog-client-go v1.28.1-0.20241225183733-80a5e1ba7a2c
 
 // replace github.com/jfrog/build-info-go => github.com/jfrog/build-info-go v1.8.9-0.20241121100855-e7a75ceee2bd
 
 // replace github.com/jfrog/gofrog => github.com/jfrog/gofrog v1.3.3-0.20231223133729-ef57bd08cedc
+
+replace github.com/jfrog/jfrog-cli-artifactory => github.com/bhanurp/jfrog-cli-artifactory v0.1.12-0.20250110100951-d002779f4509
```
RobiNino, line 105: This is probably redundant now :)

RobiNino (approved):
small comment, otherwise LGTM!

go.mod
bhanurp: Removed the replace statement

PR #1351, opened 2025-02-18
Migrate life cycle commands from jfrog-cli to jfrog-cli-artifactory

bhanurp:
-----
- Moved some common functions used across commands from cli to cli-core


common/cliutils/utils.go
```diff
@@ -251,3 +253,9 @@ func FixWinPathsForFileSystemSourcedCmds(uploadSpec *spec.SpecFiles, specFlag, e
 		}
 	}
 }
+
+func PlatformToLifecycleUrls(lcDetails *coreConfig.ServerDetails) {
```
RobiNino, line 257: This method isn't really a common util, it can probably be based in a lifecycle package. Maybe even in jfrog-cli-artifactory.

plugins/components/commandcomp.go
```diff
@@ -183,6 +206,18 @@ func WithBoolDefaultValueFalse() BoolFlagOption {
 	}
 }
 
+func WithHiddenTrue() StringFlagOption {
```
RobiNino, line 209: A similar func already exists - `SetHiddenStrFlag`

plugins/common/utils.go
```diff
@@ -124,3 +134,71 @@ func getCiValue() bool {
...
+
+// Return argument if not empty or retrieve from environment variable
+func getOrDefaultEnv(arg, envKey string) string {
+	if arg != "" {
+		return arg
+	}
+	return os.Getenv(envKey)
+}
+
+func CreateDownloadConfiguration(c *components.Context) (downloadConfiguration *artifactoryUtils.DownloadConfiguration, err error) {
```
RobiNino, line 152: Some of these aren't really plugins utils, I suggest placing them in jfrog-cli-artifactory.
In addition, since they duplicate existing functions in `jfrog-cli` (only differ in context), let's try to make a common function out of them or somehow overcome the context difference

RobiNino (changes requested):

plugins/common/utils.go
bhanurp: Agree moved lifecycle functions to `jfrog-cli-artifactory`

common/cliutils/utils.go
bhanurp: agree moved to `jfrog-cli-artifactory`

RobiNino: Looks like this is still here, did it move?

RobiNino (changes requested):

common/cliutils/utils.go
bhanurp: It didn't move earlier, updated now

RobiNino (approved):

PR #1356, opened 2025-02-24
Fix for #2874 release-bundle-create is not working with projects

bhanurp:
-----

### DIFF
--- a/common/spec/specfiles_test.go
+++ b/common/spec/specfiles_test.go
@@ -7,23 +7,31 @@ import (
 
 func TestCreateSpecFromBuildNameAndNumber(t *testing.T) {
 	t.Run("Valid Inputs", func(t *testing.T) {
-		spec, err := CreateSpecFromBuildNameAndNumber("Common-builds", "1.2.0")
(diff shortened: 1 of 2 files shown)


common/spec/specfiles_test.go
```diff
@@ -7,31 +7,39 @@ import (
 
 func TestCreateSpecFromBuildNameAndNumber(t *testing.T) {
 	t.Run("Valid Inputs", func(t *testing.T) {
-		spec, err := CreateSpecFromBuildNameAndNumber("Common-builds", "1.2.0")
+		spec, err := CreateSpecFromBuildNameAndNumber("Common-builds", "1.2.0", "test")
```
EyalDelarea, line 10: Let's extract the logic of projects key into a different test function
or we should change the name to `TestCreateSpecFromBuildNameAndNumberAndProjectKey`, which seems a bit long :)

EyalDelarea (changes requested):

common/spec/specfiles_test.go
bhanurp: Updated as `TestCreateSpecFromBuildNameNumberAndProjectKey`

PR #1373, opened 2025-04-09
Support multiple image tags in build-docker-create using image-file

agrasth:
-----

### DIFF
--- a/artifactory/utils/container/image.go
+++ b/artifactory/utils/container/image.go
@@ -169,3 +169,7 @@ func buildRequestUrl(longImageName, imageTag, containerRegistryUrl string, https
 	}
 	return "http://" + endpoint
 }
+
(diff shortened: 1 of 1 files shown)


artifactory/utils/container/image.go
```diff
@@ -169,3 +169,7 @@ func buildRequestUrl(longImageName, imageTag, containerRegistryUrl string, https
 	}
 	return "http://" + endpoint
 }
+
+func (i *Image) SetName(name string) {
```
bhanurp, line 173: This is not required if we can use NewImage() for every new tag

bhanurp (changes requested):
Try if we can avoid this new setter

agrasth:
Implemented [this](https://github.com/jfrog/jfrog-cli-artifactory/pull/61) without this setter SetName

PR #1540, opened 2026-03-31
Fix reference token privilege escalation

reshmifrog:
- [ ] This pull request is on the master branch.
-----
Problem: When a user configures JFrog CLI with a restricted reference token (e.g Viewer scope) as the password, the CLI automatically creates a new token using the deprecated Artifactory API (POST /api/security/token) with scope=member-of-groups:*. This deprecated API does not enforce scope ceiling, so it issues a fully privileged token escalating from read-only to full delete permissions.
Solution: Migrated the CLI's token creation from the deprecated Artifactory API to the current Access API (POST /access/api/v1/tokens) with scope=applied-permissions/user. The Access API enforces server-side scope ceiling if the authenticating credential has restricted scope, the server rejects the request with 403 Forbidden, preventing privilege escalation at its root.

### DIFF
--- a/utils/config/tokenrefresh.go
+++ b/utils/config/tokenrefresh.go
@@ -208,7 +208,7 @@ func createTokensForConfig(serverDetails *ServerDetails, expirySeconds int) (aut
 	// If basic auth failed and a password is available, retry using it as a Bearer token.
 	// This handles reference tokens, which the Access service can resolve server-side.
 	if serverDetails.Password != "" {
-		bearerDetails := copyServerDetailsWithAccessToken(serverDetails, serverDetails.Password)
+		bearerDetails := serverDetailsForBearerAuth(serverDetails, serverDetails.Password)
 		servicesManager, err = createAccessTokensServiceManager(bearerDetails)
 		if err != nil {
 			return auth.CreateTokenResponseData{}, err
@@ -227,15 +227,14 @@ func createTokensForConfig(serverDetails *ServerDetails, expirySeconds int) (aut
 		err.Error())
 }
 
-// copyServerDetailsWithAccessToken creates a shallow copy of ServerDetails with the given
-// token set as AccessToken and credentials cleared, so that the Access API uses Bearer
-// authentication instead of basic auth.
-func copyServerDetailsWithAccessToken(original *ServerDetails, token string) *ServerDetails {
-	copy := *original
-	copy.AccessToken = token
-	copy.User = ""
-	copy.Password = ""
-	return &copy
+// serverDetailsForBearerAuth returns a ServerDetails configured to authenticate using the
+// given token as a Bearer token, with user/password credentials cleared.
+func serverDetailsForBearerAuth(original *ServerDetails, token string) *ServerDetails {
+	details := *original
+	details.AccessToken = token
+	details.User = ""
+	details.Password = ""
+	return &details
 }
 
 func CreateInitialRefreshableTokensIfNeeded(serverDetails *ServerDetails) (err error) {
--- a/utils/config/tokenrefresh.go
+++ b/utils/config/tokenrefresh.go
@@ -2,12 +2,12 @@ package config
 
 import (
 	"errors"
-	"strings"
 	"sync"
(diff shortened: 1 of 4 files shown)


fluxxBot (approved):

utils/config/tokenrefresh.go
```diff
@@ -183,19 +184,22 @@ func writeNewTokens(serverConfiguration *ServerDetails, serverId, accessToken, r
 }
 
 func createTokensForConfig(serverDetails *ServerDetails, expirySeconds int) (auth.CreateTokenResponseData, error) {
-	servicesManager, err := createArtifactoryTokensServiceManager(serverDetails)
+	servicesManager, err := createAccessTokensServiceManager(serverDetails)
```
itsmeleela: What will happen to backward compatibility? This seems a new API introduced sometime back - will now get failures where they previously had (insecure) success. Is there a version check or a fallback path?

utils/config/tokenrefresh.go
```diff
@@ -183,19 +184,22 @@ func writeNewTokens(serverConfiguration *ServerDetails, serverId, accessToken, r
...
-	createTokenParams.ExpiresIn = expirySeconds
-	// User-scoped token
-	createTokenParams.Scope = "member-of-groups:*"
-	createTokenParams.Refreshable = true
+	expiresIn := uint(max(expirySeconds, 0)) // #nosec G115 -- expirySeconds is validated positive by callers
+	createTokenParams := accessservices.CreateTokenParams{
+		CommonTokenParams: auth.CommonTokenParams{
+			Scope:       "applied-permissions/user",
+			ExpiresIn:   &expiresIn,
+			Refreshable: clientutils.Pointer(true),
```
itsmeleela, line 193: If the Artifactory instance has refresh tokens disabled globally, this call might fail or return a token without a refresh token right? Did we validate that behavior? I dont remember if we still have an option of disabling refresh tokens

utils/config/tokenrefresh.go
```diff
@@ -183,19 +184,22 @@ func writeNewTokens(serverConfiguration *ServerDetails, serverId, accessToken, r
...
 	}
 
-	createTokenParams := services.NewCreateTokenParams()
-	createTokenParams.Username = serverDetails.User
-	createTokenParams.ExpiresIn = expirySeconds
-	// User-scoped token
-	createTokenParams.Scope = "member-of-groups:*"
-	createTokenParams.Refreshable = true
+	expiresIn := uint(max(expirySeconds, 0)) // #nosec G115 -- expirySeconds is validated positive by callers
+	createTokenParams := accessservices.CreateTokenParams{
```
itsmeleela, line 189: What will happen if this call fails with 403?  What does user see? 

Consider whether createTokensForConfig or its caller should detect a 403 and surface a message like: "Token creation failed, your credentials may have insufficient permissions to generate a new token." 

Right now there's no indication of this in the PR

utils/config/tokenrefresh.go
reshmifrog: The Access API has been available since JFrog Platform 7.x. Rather than falling back to the deprecated Artifactory API (which has the security vulnerability this PR fixes), we surface a clear error asking the user to upgrade or use --basic-auth-only. Falling back to the old API would reintroduce the privilege escalation bug.

utils/config/tokenrefresh.go
reshmifrog: Now the user now sees a descriptive message with two actionable options either upgrade their platform or use --basic-auth-only.

utils/config/tokenrefresh.go
reshmifrog: When allow-refreshable is set to false globally, the Access API does not reject the request it silently creates the token as non-refreshable (returns an empty refresh_token). When the token later expires, the refresh attempt fails, and the CLI falls back to creating a new token via createTokensForConfig. The CLI continues to work normally, just creating new tokens instead of refreshing.

itsmeleela (approved):

PR #1542, opened 2026-04-15
RTECO-992 - Added support for uv package manager

agrasth:
## Summary

Adds `UV` to the `ProjectType` enum and `"uv"` to `ProjectTypes` to register the UV package manager as a known project type.

> **Note:** `project.UV` is intentionally not added — the native UV implementation routes directly through `NativeUVCommand` and does not use the `pythonCmd` routing path, so the constant is not required.

### DIFF
--- a/common/project/projectconfig.go
+++ b/common/project/projectconfig.go
@@ -48,6 +48,7 @@ const (
 	Helm
 	Ruby
 	Conan
+	UV
 )
 
 type ConfigType string
@@ -79,6 +80,7 @@ var ProjectTypes = []string{
 	"helm",
 	"ruby",
 	"conan",
+	"uv",
 }
 
 func (projectType ProjectType) String() string {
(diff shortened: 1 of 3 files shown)


common/project/projectconfig.go
```diff
@@ -79,6 +80,7 @@ var ProjectTypes = []string{
 	"helm",
 	"ruby",
 	"conan",
+	"uv",
```
bhanurp: Why is this change needed?

agrasth: You're right — this change is not needed. The native UV implementation creates `NativeUVCommand` directly and never routes through `pythonCmd`, so `project.UV` was never referenced anywhere. Removed both the iota constant and the `"uv"` string from `ProjectTypes`. If a config-file-based `jf rt python uv` path is added in the future, it can be re-added then.

naveenku-jfrog (approved):

PR #1560, opened 2026-05-19
RTECO-1266 - Add Subcommand to command

udaykb2:
We can now pass subcommand to a command so
commands like "jf ai plugins publish" work

- [x] This pull request is on the master branch.
-----


plugins/components/conversionlayer.go
```diff
@@ -99,6 +99,13 @@ func convertCommand(cmd Command, namespaces ...string) (cli.Command, error) {
 		SkipFlagParsing: cmd.SkipFlagParsing,
 		Hidden:          cmd.Hidden,
 	}
+	if len(cmd.Subcommands) > 0 {
```
bhanurp, line 105: A good test scenario to consider what happens if SkipFlagParsing=true and there are len(SubCommands) > 0.

bhanurp (approved):

bhanurp (approved):

plugins/components/conversionlayer.go
udaykb2: updated, added check

naveenku-jfrog (approved):

PR #1613, opened 2026-08-19
Resolve AI client from host app and tighten session detectors

sverdlov93:
## Overview

CLI telemetry stamps `client` with the real host app of a proven agent session: known IDE first, then Claude Code as a standalone app, then a short terminal name. Session detection is tightened so grok, kilo, Claude cowork, Copilot, and alias names match product rules.

## Details

Host (`ai-client/` / JSON `client`), strongest signal first:

- Known IDE: `cursor`, `vscode`, `zed`, `jetbrains`, `windsurf`, `antigravity`, `codium`, `trae`, `visualstudio` from editor-owned env (never raw `TERM_PROGRAM` for forks).
- Cursor / Windsurf / Antigravity / Trae / VSCodium keep their name even when `TERM_PROGRAM=vscode`.
- Askpass matching uses the editor install (`/cursor.app`, `/cursor/resources`, or `/.cursor-server` for Remote-SSH), not a username in the path.
- Desktop Visual Studio (`VisualStudioVersion`) is not labelled `vscode`.
- Claude Code with no proven IDE is `claude`, not iTerm.
- Otherwise map `TERM_PROGRAM` to a short product name: `iTerm.app` → `iterm`, `WarpTerminal` → `warp`, `Apple_Terminal` → `terminal`, `tmux` → `tmux`.
- When `TERM_PROGRAM` is empty or inherited `vscode`, fall back to `TMUX`, `WT_SESSION` (`windows-terminal`), Ghostty, Kitty, or Alacritty.
- Generic `GIT_ASKPASS` is not used as IDE proof.
- Humans still get no client; the axis is gated on an agent session.

Session (`ai-agent/`):

- Kilo only matches `KILOCODE_FEATURE=cli`. Grok matches exact `GROK_AGENT=1`. Cowork is classified as Claude before the generic Claude detector.
- Copilot extras: `COPILOT_AGENT`, `COPILOT_AGENT_JOB_ID`. Aliases such as `github_copilot_vscode_agent` and `grok-cli` map to canonical session names. The vscode-plugin alias is folded the same way as session detection (`AI_AGENT`/`AGENT`, case and `@version`) and last-resorts `client` to `vscode` when no IDE is proven.

```mermaid
flowchart LR
  env[Process env] --> detect[Detect agent session]
  detect --> host[Resolve host app]
  host --> ide[Known IDE]
  ide --> app[Claude if no IDE]
  app --> term[Short terminal name]
  term --> ua["ai-agent/ and ai-client/ on User-Agent"]
  ua --> metrics[Coralogix metrics]
```

## Notes

Companion PRs: jfrog-cli tests, jfrog-agent-hooks stamp, jfrog-skills EXTRACT split. Merge this module first so CLI can bump `jfrog-cli-core`.

### DIFF
--- a/common/commands/execution_context.go
+++ b/common/commands/execution_context.go
@@ -222,7 +222,7 @@ func detectModel() string {
 // | cursor      | CURSOR_TRACE_ID, agent=cursor, or "cursor" askpass path    |
 // | windsurf    | "windsurf" askpass path                                    |
 // | antigravity | "antigravity" askpass path                                 |
-// | vscode      | first-party Copilot plugin markers                         |
+// | vscode      | Copilot plugin markers, or stock VS Code askpass           |
 //
 // Order is significant: every VS Code fork inherits TERM_PROGRAM=vscode and the
 // VSCODE_* vars from upstream, so forks must resolve before anything reports
@@ -247,6 +247,9 @@ func detectClient(agent string) string {
 	case os.Getenv("COPILOT_AGENT") == "1", os.Getenv("AI_AGENT") == "github_copilot_vscode_agent":
 		// Last resort: these prove the first-party plugin, not the window. A
 		// JetBrains host is caught above, so this only fires with no host marker.
+		return "vscode"
+	case os.Getenv("VSCODE_GIT_ASKPASS_MAIN") != "", os.Getenv("VSCODE_GIT_ASKPASS_NODE") != "":
+		// Stock VS Code after forks were ruled out. GIT_ASKPASS is generic git.
 		return "vscode"
 	default:
 		return ""
--- a/common/commands/execution_context.go
+++ b/common/commands/execution_context.go
@@ -23,8 +23,10 @@ type ExecutionContext struct {
 	IsInteractive bool   // stdout is a TTY
 	TraceID       string // e.g. CURSOR_TRACE_ID; empty if none
 
-	// Client: IDE host only — "cursor" or "vscode". Empty when there is no IDE
-	// (Claude Code TUI, Copilot CLI, Gemini, …). Never TERM_PROGRAM / a terminal emulator.
+	// Client: host editor window — "cursor", "vscode", "zed", "jetbrains",
+	// "windsurf", "antigravity". Empty when the session has no editor (Claude
+	// Code TUI, Copilot CLI, Gemini, CI) or when the host cannot be proven.
+	// Never TERM_PROGRAM / a terminal emulator.
 	Client string
 
 	// Model: model slug (JFROG_CLI_AI_MODEL) — "opus-4.7".
@@ -74,7 +76,7 @@ type agentDetector struct {
 // | amazon_q    | AWS_EXECUTION_ENV contains AmazonQ-For-CLI                   |
 // | unknown     | AI_AGENT/AGENT set to an unrecognized value                  |
 //
-// Client axis: IDE only (cursor / vscode). Never TERM_PROGRAM.
+// Client axis: host editor window, from editor-owned env — see detectClient.
 // Model axis: sanitized JFROG_CLI_AI_MODEL (skill/user supplied slug).
 var agentEnvDetectors = []agentDetector{
 	// GROK_AGENT is also a profile *path* for humans — exact "1" only.
@@ -208,22 +210,62 @@ func detectModel() string {
 	return sanitizeToken(os.Getenv("JFROG_CLI_AI_MODEL"))
 }
 
-// detectClient returns the IDE host for an agent session. Cursor always
-// reports "cursor" (agent shells often inherit TERM_PROGRAM=vscode). Copilot
-// reports "vscode" only for the first-party VS Code plugin. Every other
-// agent — including Copilot CLI and Claude — omits the axis.
+// detectClient returns the editor window hosting an agent session. The host is
+// read from env the *editor* owns, not inferred from the agent name: Copilot and
+// Claude both run inside JetBrains, Zed and VS Code, so agent alone cannot name
+// the window.
+//
+// | Client      | Host-owned signal                                          |
+// |-------------|------------------------------------------------------------|
+// | zed         | ZED_TERM                                                   |
+// | jetbrains   | TERMINAL_EMULATOR=JetBrains-JediTerm                       |
+// | cursor      | CURSOR_TRACE_ID, agent=cursor, or "cursor" askpass path    |
+// | windsurf    | "windsurf" askpass path                                    |
+// | antigravity | "antigravity" askpass path                                 |
+// | vscode      | first-party Copilot plugin markers                         |
+//
+// Order is significant: every VS Code fork inherits TERM_PROGRAM=vscode and the
+// VSCODE_* vars from upstream, so forks must resolve before anything reports
+// "vscode" (product rule P13 — a Cursor session must never be labelled vscode).
+// Unproven hosts omit the axis rather than guess.
+//
+// JetBrains publishes one terminal value for the whole family, so IntelliJ,
+// GoLand and PyCharm collapse to "jetbrains"; separating them would need a
+// parent-process walk on every command.
 func detectClient(agent string) string {
-	switch agent {
-	case "cursor":
+	switch {
+	case os.Getenv("ZED_TERM") != "":
+		return "zed"
+	case os.Getenv("TERMINAL_EMULATOR") == "JetBrains-JediTerm":
+		return "jetbrains"
+	case agent == "cursor", os.Getenv("CURSOR_TRACE_ID") != "", askpassPathContains("cursor"):
 		return "cursor"
-	case "copilot":
-		if os.Getenv("COPILOT_AGENT") == "1" || os.Getenv("AI_AGENT") == "github_copilot_vscode_agent" {
-			return "vscode"
-		}
-		return ""
+	case askpassPathContains("windsurf"):
+		return "windsurf"
+	case askpassPathContains("antigravity"):
+		return "antigravity"
+	case os.Getenv("COPILOT_AGENT") == "1", os.Getenv("AI_AGENT") == "github_copilot_vscode_agent":
+		// Last resort: these prove the first-party plugin, not the window. A
+		// JetBrains host is caught above, so this only fires with no host marker.
+		return "vscode"
 	default:
 		return ""
 	}
+}
+
+// askpassEnvVars hold the path of the editor's git askpass helper, which embeds
+// the application name (…/Cursor.app/…, …/windsurf/…). It is the only env that
+// separates a VS Code fork from upstream, since forks copy the VSCODE_* names
+// verbatim. Path-shaped, so match case-insensitively on a substring.
+var askpassEnvVars = []string{"VSCODE_GIT_ASKPASS_MAIN", "VSCODE_GIT_ASKPASS_NODE", "GIT_ASKPASS"}
+
+func askpassPathContains(app string) bool {
+	for _, env := range askpassEnvVars {
+		if strings.Contains(strings.ToLower(os.Getenv(env)), app) {
+			return true
+		}
+	}
+	return false
 }
 
 // maxTokenLen caps sanitized identity tokens so a pathological env value cannot
--- a/common/commands/execution_context.go
+++ b/common/commands/execution_context.go
@@ -185,6 +185,8 @@ var agentEnvDetectors = []agentDetector{
 	{"iflow", []string{"IFLOW_CLI"}, nil, nil},
 	{nameTrae, []string{"TRAE_AI_SHELL_ID"}, nil, nil},
 	{"pi", []string{"PI_CODING_AGENT"}, nil, nil},
+	// Amazon Q Developer CLI sets AWS_EXECUTION_ENV to a value containing
+	// AmazonQ-For-CLI (AWS CLI execution-environment convention).
 	{nameAmazonQ, nil, nil, map[string]string{"AWS_EXECUTION_ENV": "AmazonQ-For-CLI"}},
 }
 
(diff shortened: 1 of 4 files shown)


common/commands/execution_context_test.go
```diff
@@ -256,23 +318,193 @@ func TestSanitizeToken(t *testing.T) {
...
+		{"claude in vscode via askpass", map[string]string{"CLAUDE_CODE_CHILD_SESSION": "1", EnvVSCodeGitAskpassMain: vscodeGitHelperPath("/Applications/Visual Studio Code.app/Contents/Resources/app/extensions/git/dist")}, NameVSCode},
+		{"windsurf agent without askpass", map[string]string{"WINDSURF_CASCADE_TERMINAL": "1", EnvTermProgram: NameVSCode}, NameWindsurf},
+		{"antigravity agent without askpass", map[string]string{"ANTIGRAVITY_AGENT": "1", EnvTermProgram: NameVSCode}, NameAntigravity},
+		// Other CLI agents fall back to the terminal app.
+		{"gemini in iterm", map[string]string{"GEMINI_CLI": "1", EnvTermProgram: TermProgramItermApp}, NameIterm},
+		{"gemini in warp", map[string]string{"GEMINI_CLI": "1", EnvTermProgram: TermProgramWarp}, NameWarp},
+		{"gemini in apple terminal", map[string]string{"GEMINI_CLI": "1", EnvTermProgram: TermProgramApple}, NameTerminal},
+		{"gemini in tmux", map[string]string{"GEMINI_CLI": "1", EnvTermProgram: NameTmux}, NameTmux},
+		{"generic git askpass does not claim cursor", map[string]string{"GEMINI_CLI": "1", EnvTermProgram: TermProgramItermApp, EnvGitAskpass: "/Users/cursor/bin/git-helper"}, NameIterm},
```
sankerr: 🤖 **Go-Sec is still failing on this head — this is the one remaining G101.**

[Run 32963839166](https://github.com/jfrog/jfrog-cli-core/actions/runs/32963839166/job/98161884153):

```
[common/commands/execution_context_test.go:452] - G101 (CWE-798): Potential hardcoded credentials (Confidence: LOW, Severity: HIGH)
```

The const-block suppressions and the `vscodeGitHelperPath` / `vscodeGitHelperPathWin` helpers fixed the other four hits, but this case still assigns a string literal to a key whose name contains `pass` (`EnvGitAskpass`), and the `// #nosec G101 -- fixture path, not a credential` it carried at the previous head was dropped when the table was reformatted.

Minimal fix: restore the trailing annotation on this line, or route the value through a small helper the way the askpass fixtures now do.

common/commands/execution_context.go
```diff
@@ -181,18 +263,163 @@ func computeExecutionContext() ExecutionContext {
...
+// GoLand and PyCharm collapse to "jetbrains"; separating them would need a
+// parent-process walk on every command. Do not split the family via macOS
+// bundle IDs (peers that do that also mis-label JediTerm as PyCharm).
+func detectClient(agent string) string {
+	switch {
+	case os.Getenv(EnvZedTerm) != "":
+		return NameZed
+	case os.Getenv(EnvTerminalEmulator) == JediTerm:
+		return NameJetBrains
+	case agent == NameCursor, os.Getenv(EnvCursorTraceID) != "", askpassPathContains(NameCursor):
```
sankerr: 🤖 **`agent == NameCursor` can only change the result when the host is *not* Cursor.**

A genuine Cursor window is already proven by the other two disjuncts on this line: `CURSOR_TRACE_ID` (which the detector table documents as set for Cursor integrated terminals) or the askpass path, which now also covers `.cursor-server` for remote SSH. So `agent == NameCursor` only decides the outcome when neither holds — and because this case sits above the Visual Studio, VSCodium and stock-VS-Code cases, it wins over harder evidence:

- `cursor-agent` run from a stock VS Code terminal (`VSCODE_GIT_ASKPASS_MAIN` → `Visual Studio Code.app`) reports `client=cursor`.
- `cursor-agent` in iTerm, Warp, or headless CI reports `client=cursor`, discarding the terminal.

That contradicts the axis this PR defines, and the comment on `TestDetectExecutionContext_ClientFollowsHostEditor` ("The host axis must follow the editor, not the agent"). The `windsurf` / `antigravity` / `trae` equivalents below are fine, because their markers only exist inside those IDEs' terminals; `CURSOR_AGENT` is the one marker also set by the standalone CLI, so agent identity does not imply the window.

Dropping the `agent == NameCursor` disjunct leaves genuine Cursor sessions correctly labelled by trace ID or askpass. No test covers the mislabel today.

common/commands/execution_context.go
```diff
@@ -181,18 +263,163 @@ func computeExecutionContext() ExecutionContext {
...
+		return NameWindsurf
+	case agent == NameAntigravity, askpassPathContains(NameAntigravity):
+		return NameAntigravity
+	case agent == NameTrae, askpassPathContains(NameTrae):
+		return NameTrae
+	case askpassPathContains(AskpassVSCodium), askpassPathContains(NameCodium):
+		return NameCodium
+	case os.Getenv(EnvVisualStudioVersion) != "":
+		return NameVisualStudio
+	case os.Getenv(EnvCopilotAgent) == "1", isCopilotVSCodePluginAlias():
```
sankerr: 🤖 **This fires even when `TERM_PROGRAM` names a concrete terminal, contrary to the comment.**

The comment says it "only fires with no host marker", but terminal resolution happens in `default:` well below, so a definitive terminal signal loses to this plugin guess. Two tests in this file, differing only in which Copilot variable is set with identical host env, lock the inconsistency in:

- `TestDetectExecutionContext_ClientCopilotVscodePlugin`: `COPILOT_AGENT=1` + `TERM_PROGRAM=iTerm.app` → `vscode`
- `TestDetectExecutionContext_ClientCopilotCLIFallsBackToTerminalApp`: `COPILOT_CLI=1` + the same `TERM_PROGRAM` → `iterm`

The same branch also labels the headless Copilot coding agent as `vscode` — its sibling marker in the detector table above is `COPILOT_AGENT_JOB_ID`, i.e. a CI job with no editor at all.

Keeping `vscode` as the *no-evidence* fallback is reasonable; it should lose to a known `TERM_PROGRAM` value. Moving this case below terminal resolution (or gating it on `canonicalTerminalName(TERM_PROGRAM) == ""`) fixes both while still catching an extension-host child that has no terminal env.

common/commands/execution_context.go
```diff
@@ -181,18 +263,163 @@ func computeExecutionContext() ExecutionContext {
...
+// Remote-SSH installs use ~/.cursor-server (and the same .*-server layout
+// on other forks); those must not fall through to client=vscode.
+func askpassPathContains(app string) bool {
+	needle := strings.ToLower(app)
+	for _, env := range askpassEnvVars {
+		p := strings.ToLower(strings.ReplaceAll(os.Getenv(env), "\\", "/"))
+		if p == "" {
+			continue
+		}
+		if strings.Contains(p, "/"+needle+".app") ||
```
sankerr, line 423: 🤖 **Fork matching fails open into `vscode`, which is the P13 violation this function exists to prevent.**

All three patterns require the app name to be a whole path segment (`/<app>.app`, `/<app>/resources`, `/.<app>-server`), so a fork whose install directory carries a suffix misses every fork case — `Cursor Nightly.app`, `Windsurf Next.app`, or any fork not in the hardcoded list. Those then fall through to the stock-VS-Code case below, which only checks that the askpass var is non-empty, so a Cursor session gets labelled `vscode`.

Suggest making that last branch positively identify a VS Code install (`/visual studio code`, `/microsoft vs code`, `/code/resources`, `/.vscode-server`) and return `""` otherwise, so an unrecognized fork fails safe instead of being attributed to upstream. There is no test for a suffixed install path today.

common/commands/execution_context.go
```diff
@@ -181,18 +263,163 @@ func computeExecutionContext() ExecutionContext {
...
+	name = strings.TrimSuffix(name, ".app")
+	if mapped, ok := terminalNameAliases[name]; ok {
+		return mapped
+	}
+	// TERM_PROGRAM=vscode is inherited by every VS Code fork. Do not treat it
+	// as a proven vscode window — that is how Copilot CLI gets mislabelled.
+	if name == NameVSCode {
+		return ""
+	}
+	return name
```
sankerr: 🤖 **The allowlist introduced just above is bypassed for anything unrecognized.**

`TERM_PROGRAM=FooBar.app` becomes `client=foobar` (asserted in the host table as "gemini in unknown terminal app"), up to 64 chars, and reaches a Coralogix label verbatim via `commands_count_metric.go`. The agent axis solves exactly this with `AgentUnknown`.

The passthrough predates this PR, but this PR is what establishes the allowlist that it now contradicts — returning `""` or a fixed `unknown` for unmapped values would keep the `client` label bounded and consistent with `agent`.

common/commands/execution_context.go
```diff
@@ -64,63 +135,74 @@ type agentDetector struct {
...
+	{"opencode", []string{"OPENCODE", "OPENCODE_SESSION_ID"}, nil, nil},
+	{"amp", []string{"AMP_CURRENT_THREAD_ID"}, nil, nil},
+	{"augment", []string{"AUGMENT_AGENT"}, nil, nil},
+	{NameQwen, []string{"QWEN_CODE"}, nil, nil},
+	{NameAntigravity, []string{"ANTIGRAVITY_AGENT"}, nil, nil},
+	{"crush", []string{"CRUSH"}, nil, nil},
+	{"iflow", []string{"IFLOW_CLI"}, nil, nil},
+	{NameTrae, []string{"TRAE_AI_SHELL_ID"}, nil, nil},
+	{"pi", []string{"PI_CODING_AGENT"}, nil, nil},
+	{NameAmazonQ, nil, nil, map[string]string{"AWS_EXECUTION_ENV": "AmazonQ-For-CLI"}},
```
sankerr: 🤖 **The only detector without a documented source, and the table test cannot catch a wrong literal.**

`TestDetectAgent_FromTable` builds the fixture from the table it asserts (`t.Setenv(k, "prefix-"+substr+"-suffix")`), so it passes for any value. If the real `AWS_EXECUTION_ENV` spelling or casing differs from `AmazonQ-For-CLI`, `strings.Contains` never matches, Amazon Q sessions silently fall back to the generic `AI_AGENT` path, and CI stays green.

Every neighbouring entry records why its marker is trustworthy; the same one-line citation here (and for `KILOCODE_FEATURE=cli` / `GROK_AGENT=1`, verified against a live session) would make the literal reviewable.

common/commands/execution_context.go
```diff
@@ -13,6 +13,68 @@ import (
 // metric cardinality bounded.
 const AgentUnknown = "unknown"
 
+// Canonical wire names reused as detector names, alias targets, and host-client
+// results. One-off table keys stay string literals.
+const (
```
sankerr, line 19: 🤖 **~40 new exported identifiers in a public package of a versioned module.**

`NameTerminal`, `NameTmux`, `EnvTerm`, `EnvAgent`, `JediTerm`, `TermProgramApple`, `AskpassVSCodium` and friends become public API of `common/commands` that cannot be renamed without a breaking change, for what is internal detection detail. Same-package tests do not need exported names.

Two concrete costs: `EnvGitAskpass` is referenced only by a test fixture and a prose comment, never by production code; and turning the askpass env names into declarations is what required the three `#nosec G101` suppressions in this block (inline literals at the use sites did not trip the rule).

sankerr (commented):
🤖 Review comment by Cursor AI

Reviewed at `065e2a6e`. Host independence, fail-open vscode, exported constants, unknown TERM_PROGRAM cardinality, and the Amazon Q marker citation are addressed.

### Findings
✅ All findings have been resolved.

⚠️ AI review supplements human review — always get a human reviewer too.

common/commands/execution_context.go
sverdlov93: Fixed — Copilot plugin markers now lose to a known terminal (`COPILOT_AGENT=1` + iTerm → iterm). vscode remains only when no terminal is proven.

common/commands/execution_context.go
sverdlov93: Fixed — dropped `agent == cursor`. Cursor is only proven by CURSOR_TRACE_ID or askpass. CURSOR_AGENT in iTerm or stock VS Code now follows that window. Tests added.

common/commands/execution_context_test.go
sverdlov93: Fixed — restored `#nosec G101 jfrog-ignore` on this GIT_ASKPASS fixture in c5f82c37.

common/commands/execution_context.go
sverdlov93: Fixed — unexported the Name/Env/askpass/terminal constants. AgentUnknown stays public. Same-package tests use the unexported names.

common/commands/execution_context.go
sverdlov93: Fixed — stock vscode now requires a positive install path (`/visual studio code`, `/microsoft vs code`, `/.vscode-server`, `/code/resources`). Cursor Nightly.app no longer fails open as vscode.

common/commands/execution_context.go
sverdlov93: Fixed — unmapped TERM_PROGRAM (including FooBar.app and inherited vscode) now returns empty. Client stays on the known-app allowlist.

common/commands/execution_context.go
sverdlov93: Fixed — cited AWS_EXECUTION_ENV / AmazonQ-For-CLI on the detector row, and added TestDetectAgent_AmazonQExecutionEnv with a hardcoded literal so a table typo fails CI.

sankerr (approved):

lesnerd (approved):

omerzi (approved):

